PR #27, opened 2024-10-08
Endpoint pages

RalkeyOfficial:


src/store/modules/endpoints.js
```diff
@@ -1,129 +1,122 @@
...
+			}
+
+			// convert to an entity
+			endpointItem = new Endpoint(endpointItem)
+
+			// verify data with Zod
+			const validationResult = endpointItem.validate()
+			if (!validationResult.success) {
+				console.error(validationResult.error)
+				console.log(validationResult)
```
remko48, line 88: There is a console.log().

src/store/modules/endpoints.js
```diff
@@ -1,129 +1,122 @@
...
+		async saveEndpoint(endpointItem) {
+			if (!endpointItem) {
+				throw new MissingParameterError('endpointItem')
+			}
+
+			// convert to an entity
+			endpointItem = new Endpoint(endpointItem)
+
+			// verify data with Zod
+			const validationResult = endpointItem.validate()
```
remko48, line 85: When I update an endpoint I get multiple errors.

1. Id is number not string.
2. Created is not a valid date time
3. Updated is not a valid date time

These errors should not come back since id should be set as a string (which is backend can be fixed in frontend)
And Created and Updated is set in the backend and should not be checked or validated when updating

remko48 (changes requested):

PR #28, opened 2024-10-08
Added function to allow dot notation in Source->config array

WilcoLouwerse:


lib/Service/AuthenticationService.php
```diff
@@ -3,9 +3,138 @@
...
+	 *
+	 * @param array $configuration Configuration array for authentication.
+	 *
+	 * @return array|array[] The call options for OAuth with Password Credentials
+	 */
+	private function createPasswordConfig(array $configuration): array
+	{
+
+		if(($diff = array_diff(self::REQUIRED_PARAMETERS_PASSWORD, array_keys(array: $configuration))) !== []) {
+			throw new BadRequestException(message: 'Some required parameters are not set: ['.implode(separator: ',', array: $diff).']');
```
bbrands02, line 72: Can the implode be assigned to a var on an empty line above?

lib/Service/AuthenticationService.php
```diff
@@ -3,9 +3,138 @@
 namespace OCA\OpenConnector\Service;
 
 
+use GuzzleHttp\Client;
+use OAuthException;
+use Symfony\Component\HttpFoundation\Exception\BadRequestException;
 
 class AuthenticationService
```
bbrands02, line 10: Do we want class docblocks in this repo?

lib/Service/AuthenticationService.php
rjzondervan: We could try that, yes

lib/Service/AuthenticationService.php
rjzondervan: There was no docblock yet, we can add it if we want to.

lib/Service/CallService.php
```diff
@@ -169,4 +169,34 @@ public function call(
...
+				$dotConfig->set($key, $value);
+				$unsetKeys[] = $key;
+			}
+		}
+
+		// Remove the old keys containing a dot that we mapped to a different position in the $config array.
+		$config = $dotConfig->all();
+		foreach ($unsetKeys as $key) {
+			unset($config[$key]);
+		}
```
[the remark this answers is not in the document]
WilcoLouwerse: To remove these:
image

Otherwise they would still be present here:
image

lib/Service/AuthenticationService.php
WilcoLouwerse: This is not a question, we want docblocks

PR #30, opened 2024-10-10
consumer pages

RalkeyOfficial:


src/entities/consumer/consumer.ts
```diff
@@ -3,48 +3,50 @@ import { TConsumer } from './consumer.types'
 
 export class Consumer implements TConsumer {
 
-	public id: string
+	public id: number
```
remko48, line 6: When validating Zod does not like null as number.
Lets have a call to check if id should be a number here

remko48 (changes requested):

remko48 (approved):

PR #36, opened 2024-10-24
Make mappings tests available

rjzondervan:
as well as oauth and jwt tokens


lib/Controller/MappingsController.php
```diff
@@ -162,4 +164,106 @@ public function destroy(int $id): JSONResponse
...
+     * }
+     */
+    public function test(): JSONResponse
+    {
+        // Get all parameters from the request
+        $data = $this->request->getParams();
+
+
+        // Validate that required parameters are present
+        if (!isset($data['inputObject']) || !isset($data['mapping'])) {
```
bbrands02, line 211: Kan je alle checks doorlopen en === false of === true ervan maken

lib/Controller/MappingsController.php
```diff
@@ -162,4 +164,106 @@ public function destroy(int $id): JSONResponse
...
+            throw new \InvalidArgumentException('Both `inputObject` and `mapping` are required');
```
bbrands02, line 212: Denk dat we alle classes met use statements willen importen?

lib/Controller/MappingsController.php
```diff
@@ -162,4 +164,106 @@ public function destroy(int $id): JSONResponse
...
+     *     "resultObject": {
+     *         "fullName": "John Doe",
+     *         "userAge": 30,
+     *         "contactEmail": "john@example.com"
+     *     },
+     *     "isValid": true,
+     *     "validationErrors": []
+     * }
+     */
+    public function test(): JSONResponse
```
WilcoLouwerse, line 204: This function could use some cleanup

lib/Service/AuthenticationService.php
```diff
@@ -136,5 +165,99 @@ public function fetchOAuthTokens (string $endpoint, array $configuration): strin
...
+	{
+		$renderedPayload = $this->twig->createTemplate($configuration['payload'])->render($configuration);
+
+		return json_decode($renderedPayload, true);
+	}
+
+	private function getJWK(array $configuration): ?JWK
+	{
+		$jwk = null;
+		if (in_array(needle: $configuration['algorithm'], haystack: ['HS256', 'HS512'])) {
```
bbrands02, line 210: hier ook === true etc

appinfo/routes.php
```diff
@@ -19,6 +19,8 @@
 		['name' => 'endpoints#logs', 'url' => '/api/endpoints-logs/{id}', 'verb' => 'GET'],
 		['name' => 'synchronizations#contracts', 'url' => '/api/synchronizations-contracts/{id}', 'verb' => 'GET'],
 		['name' => 'synchronizations#logs', 'url' => '/api/synchronizations-logs/{id}', 'verb' => 'GET'],
+		// Mapping endpoints
+		['name' => 'mappings#test', 'url' => '/api/mappings/test', 'verb' => 'POST'],
```
WilcoLouwerse, line 23: I would expect an id here in this endpoint

lib/Service/AuthenticationService.php
```diff
@@ -3,12 +3,27 @@
...
+use Jose\Component\Signature\Algorithm\RS384;
+use Jose\Component\Signature\Algorithm\RS512;
+use Jose\Component\Signature\JWSBuilder;
+use Jose\Component\Signature\Serializer\CompactSerializer;
 use OAuthException;
 use Symfony\Component\HttpFoundation\Exception\BadRequestException;
+use Twig\Environment;
+use Twig\Loader\ArrayLoader;
 
 class AuthenticationService
```
WilcoLouwerse, line 24: This class and its functions are missing some docblocks

lib/Service/AuthenticationService.php
```diff
@@ -136,5 +165,99 @@ public function fetchOAuthTokens (string $endpoint, array $configuration): strin
...
+		$jwk = null;
+		$filename = "/srv/api/var/privatekey-$stamp";
+		try{
+			$jwk = JWKFactory::createFromKeyFile(
+				$filename,
+				null,
+				['use' => 'sig']
+			);
+		}
+		catch(\Exception $exception) {
```
WilcoLouwerse, line 181: allignment, 

} catch () (

And use import instead of \ for Exception

lib/Service/AuthenticationService.php
```diff
@@ -136,5 +165,99 @@ public function fetchOAuthTokens (string $endpoint, array $configuration): strin
 
 		return $result['access_token'];
 	}
+	private function getRSJWK(array $configuration): ?JWK
+	{
+		$stamp = microtime().getmypid();
+		file_put_contents("/srv/api/var/privatekey-$stamp", $configuration['secret']);
+		$jwk = null;
+		$filename = "/srv/api/var/privatekey-$stamp";
+		try{
```
WilcoLouwerse, line 174: try {

lib/Service/AuthenticationService.php
```diff
@@ -136,5 +165,99 @@ public function fetchOAuthTokens (string $endpoint, array $configuration): strin
...
+
+		return json_decode($renderedPayload, true);
+	}
+
+	private function getJWK(array $configuration): ?JWK
+	{
+		$jwk = null;
+		if (in_array(needle: $configuration['algorithm'], haystack: ['HS256', 'HS512'])) {
+			$jwk = $this->getHSJWK($configuration);
+		} else if (in_array(needle: $configuration['algorithm'], haystack: ['RS256', 'RS384', 'RS512'])) {
```
WilcoLouwerse, line 212: avoid else if, just use return

lib/Service/AuthenticationService.php
```diff
@@ -136,5 +165,99 @@ public function fetchOAuthTokens (string $endpoint, array $configuration): strin
...
+	{
+		$diff = array_diff(self::REQUIRED_PARAMETERS_JWT, array_keys(array: $configuration));
+		if ($diff !== []) {
+			throw new BadRequestException(message: 'Some required parameters are not set: [' . implode(separator: ',', array: $diff) . ']');
+		}
+
+		$payload = $this->getJWTPayload($configuration);
+		$jwk 	 = $this->getJWK($configuration);
+
+		if($jwk === null) {
```
WilcoLouwerse, line 256: if (

lib/Service/CallService.php
```diff
@@ -15,21 +16,53 @@
 use GuzzleHttp\Psr7\Response;
 use OCA\OpenConnector\Db\CallLog;
 use OCA\OpenConnector\Db\CallLogMapper;
+use OCA\OpenConnector\Twig\AuthenticationExtension;
+use OCA\OpenConnector\Twig\AuthenticationRuntimeLoader;
 use Symfony\Component\Uid\Uuid;
+use Twig\Environment;
+use Twig\Loader\ArrayLoader;
 
 class CallService
```
WilcoLouwerse, line 25: This class and its functions are missing some docblocks

lib/Service/CallService.php
```diff
@@ -15,21 +16,53 @@
...
+	}
+
+	private function renderValue(array|string $value, Source $source): array|string
+	{
+			if (is_array($value) === false
+				&& str_contains(haystack: $value, needle: "{{") === true
+				&& str_contains(haystack: $value, needle: "}}") === true
+			) {
+				return $this->twig->createTemplate(template: $value, name: "sourceConfig")->render(context: ['source' => $source]);
+			} else if (is_array($value) === true){
```
WilcoLouwerse, line 55: else if after a return?
just use if return, if

WilcoLouwerse (changes requested):
see comments

lib/Controller/MappingsController.php
WilcoLouwerse: Also, please do a search on !isset and !empty to replace them with === false

lib/Service/AuthenticationService.php
rjzondervan: else if is needed here because the default should be that the algorithm is not supported

PR #41, opened 2024-10-29
mapping testing part 1

RalkeyOfficial:


src/modals/MappingTest/components/TestMappingInputObject.vue
```diff
@@ -0,0 +1,81 @@
+<script setup>
+// import { navigationStore } from '../../store/store.js'
```
remko48, line 2: Commented import

src/modals/MappingTest/components/TestMappingResult.vue
```diff
@@ -0,0 +1,58 @@
...
+}
+</script>
+
+<style scoped>
+.content {
+    text-align: left;
+}
+.result {
+	margin-inline: 0.5rem !important;
+}
```
remko48, line 57: I would add the following code.
```
.result pre {
	white-space: break-spaces;
}
```

This code will prevent long sentences from going out of the div.
It however will not stop a long word from going out of the div.

src/modals/MappingTest/components/TestMappingMappingSelect.vue
```diff
@@ -0,0 +1,238 @@
...
+			this.$emit('mapping-selected', {
+				selected: event,
+			})
+		},
+		fetchMappings(currentMappingItem = null) {
+			this.mappingsLoading = true
+
+			mappingStore.refreshMappingList()
+				.then(() => {
+					if (currentMappingItem) {
```
remko48, line 125: This if statement will never fire because you set the `currentMappingItem` to `null`.
I would change it to check on `mappingStore.mappingItem`

src/modals/MappingTest/components/TestMappingResult.vue
```diff
@@ -0,0 +1,58 @@
...
+				<p>Mapping successfully tested</p>
+			</NcNoteCard>
+			<NcNoteCard v-if="mappingTest.success === false" type="error">
+				<p>Mapping failed to test</p>
+			</NcNoteCard>
+			<NcNoteCard v-if="mappingTest.error" type="error">
+				<p>{{ mappingTest.error }}</p>
+			</NcNoteCard>
+
+			<NcGuestContent class="result">
```
remko48, line 16: The component `<NcGuestContent />` should only be used as a main wrapper on a page.
And because it isn't used as such here it causes console errors

remko48 (changes requested):
Nice it works nicely, just a couple of comments

remko48 (approved):
Nice!

PR #44, opened 2024-10-31
Enable validation of mapping results with schema

rjzondervan:


lib/Controller/MappingsController.php
```diff
@@ -253,10 +271,15 @@ public function test(): JSONResponse
...
-        if ($schema !== false && $validation !== false) {
-            // TODO: Implement schema validation logic here
-            // For now, we'll just assume it's always valid
-            $isValid = true;
+        if ($schema !== false && $validation !== false && $openRegisters !== null) {
+			$result = $openRegisters->validateObject(object: $resultObject, schemaObject: $schema->getSchemaObject($urlGenerator));
+
+			$isValid = $result->isValid();
+
+			if($result->hasError() === true) {
```
WilcoLouwerse, line 279: if (

lib/Controller/MappingsController.php
```diff
@@ -253,10 +271,15 @@ public function test(): JSONResponse
...
+				// Class imported without use because it only exists when OpenRegisters is installed.
+				$validationErrors = (new \Opis\JsonSchema\Errors\ErrorFormatter())->format(error: $result->error());
```
WilcoLouwerse, line 281: Can we reach this code if openRegisters is not installed? Resulting in an exception?

lib/Service/ObjectService.php
```diff
@@ -216,4 +229,25 @@ public function aggregateObjects(array $filters, array $pipeline, array $config)
...
 
+	/**
+	 * Attempts to retrieve the OpenRegister service from the container.
+	 *
+	 * @return mixed|null The OpenRegister service if available, null otherwise.
+	 * @throws ContainerExceptionInterface|NotFoundExceptionInterface
+	 */
+	public function getOpenRegisters(): ?\OCA\OpenRegister\Service\ObjectService
+	{
+		if (in_array(needle: 'openregister', haystack: $this->appManager->getInstalledApps()) === true) {
```
WilcoLouwerse, line 240: can and should we also check version here?

WilcoLouwerse (approved):

WilcoLouwerse (approved):

PR #59, opened 2024-11-12
save object

RalkeyOfficial:


WilcoLouwerse (approved):
Approved, but someone need to test this before it gets merged!!! ⚠️

remko48 (approved):
Approved, ~~there is a backend error partially blocking the save function to work properly tho.~~

~~I'll see if I can get it fixed~~

I used an outdated version of OpenRegister while testing.
With the updated version it works like a charm!

PR #85, opened 2024-11-25
Synchronize from single-object endpoint

rjzondervan:


lib/Service/SynchronizationService.php
```diff
@@ -174,6 +175,30 @@ private function getOriginId(Synchronization $synchronization, array $object): i
         return $originId;
     }
 
+	public function getObjectFromSource(Synchronization $synchronization, string $endpoint): array
```
WilcoLouwerse, line 178: Missing docblock

lib/Service/SynchronizationService.php
```diff
@@ -383,6 +421,7 @@ public function getAllObjectsFromApi(Synchronization $synchronization, ?bool $is
         ];
 
         // Make the initial API call
+		// @TODO: method is now fixed to GET, but could end up in configuration.
```
WilcoLouwerse, line 424: Just a reminder that we should avoid doing this (even though I do it myself as well 😝)
Maybe create an issue for this, or mention it in another existing issue?

lib/Service/SynchronizationService.php
```diff
@@ -174,6 +175,30 @@ private function getOriginId(Synchronization $synchronization, array $object): i
...
+			'headers' => $headers,
+			'query' => $query,
+		];
+
+		if (str_starts_with($endpoint, $source->getLocation()) === true) {
+			$endpoint = str_replace(search: $source->getLocation(), replace: '', subject: $endpoint);
+		}
+
+		// Make the initial API call
+		// @TODO: method is now fixed to GET, but could end up in configuration.
```
WilcoLouwerse, line 196: Just a reminder that we should avoid doing this (even though I do it myself as well 😝)
Maybe create an issue for this, or mention it in another existing issue?

WilcoLouwerse (changes requested):

lib/Service/SynchronizationService.php
rjzondervan: We should create an issue for this. I refined with @rubenvdlinde that I would put in these TODOs

WilcoLouwerse: Please show me the new issue created for this

rjzondervan: https://conduction.atlassian.net/browse/CONNECTOR-152?atlOrigin=eyJpIjoiZmEwYTA0NGE4NjlkNGEzODk3NTUxNmIzYTRlMTUwZDMiLCJwIjoiaiJ9

WilcoLouwerse (approved):

PR #88, opened 2024-11-25
feature/CONNECTOR-147/job-class-dropdown

RalkeyOfficial:


src/modals/Job/EditJob.vue
```diff
@@ -28,10 +28,11 @@ import { jobStore, navigationStore } from '../../store/store.js'
 						label="Description"
 						:value.sync="jobItem.description" />
 
-					<NcTextField
-						label="Job Class"
-						maxlength="255"
-						:value.sync="jobItem.jobClass" />
+					<NcSelect v-bind="classOptions"
+						v-model="classOptions.value"
```
remko48, line 32: I would add a class called `jobClassSelect` and give it a style of `width: 100%` this will prevent the select box from shrinking and becoming unreadable

remko48 (approved):
Nice just one small comment

PR #120, opened 2024-12-05
Check if have to use pagination

bbrands02:


rjzondervan (approved):

lib/Db/Synchronization.php
```diff
@@ -65,6 +65,16 @@ public function __construct() {
 		$this->addType(fieldName:'followUps', type: 'json');
 	}
 
+    public function usesPagination(): bool
```
WilcoLouwerse, line 68: docblock?

WilcoLouwerse (changes requested):

rjzondervan (approved):

configurations/sharepoint-woo/mappings/xxllnc-suite-to-publications.json
```diff
@@ -4,23 +4,25 @@
 	"version": "",
```
WilcoLouwerse, line 4: version?

PR #123, opened 2024-12-06
Fixes on jobs and log cleanup

rjzondervan:
- Run jobs at given interval
- Store logRetention and errorRetention-fields correctly
- Store expiry of logs
- Remove expired logs

### DIFF
--- a/lib/Service/CallService.php
+++ b/lib/Service/CallService.php
@@ -138,6 +138,7 @@ class CallService
 			$callLog->setStatusCode(409);
 			$callLog->setStatusMessage("This source is not enabled");
 			$callLog->setCreated(new \DateTime());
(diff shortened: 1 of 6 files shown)


lib/Service/CallService.php
```diff
@@ -178,6 +180,7 @@ public function call(
 			$callLog->setStatusCode(429); //
 			$callLog->setStatusMessage("The rate limit for this source has been exceeded. Try again later.");
 			$callLog->setCreated(new \DateTime());
+			$callLog->setExpires(new \DateTime('now + '.$source->getErrorRetention().' seconds'));
```
bbrands02, line 183: I think this value can be defined once and re used each time

bbrands02 (approved):

PR #129, opened 2024-12-10
Added source hash mapping

bbrands02:


RalkeyOfficial (approved):

lib/Migration/Version1Date20241210100000.php
```diff
@@ -0,0 +1,62 @@
...
+		 * @var ISchemaWrapper $schema
+		 */
+		$schema = $schemaClosure();
+
+		if ($schema->hasTable('openconnector_synchronizations') === true) {
+			$table = $schema->getTable('openconnector_synchronizations');
+
+			if ($table->hasColumn('sourceHashMapping') === false) {
+				$table->dropColumn('sourceHashMapping');
+				$table->addColumn('source_hash_mapping', Types::STRING)->setNotnull(false);
```
WilcoLouwerse, line 48: This looks weird, please explain why you are doing this

WilcoLouwerse (changes requested):

lib/Migration/Version1Date20241210100000.php
bbrands02: We need to be able to map before hashing

PR #179, opened 2025-01-15
many bug fixes

Sudo-Thijn:


bbrands02 (approved):

src/modals/Synchronization/RunSynchronization.vue
```diff
@@ -1,6 +1,5 @@
 <script setup>
```
remko48, line 1: Now we can test a synchronization via the run button, but we still have a test button in the action form

remko48: And can we discuss all the other frontend changes?

remko48 (changes requested):

rjzondervan (approved):

remko48 (approved):

PR #196, opened 2025-01-29
Fetch and write files from other sources, upload file parts

rjzondervan:


lib/Service/EndpointService.php
```diff
@@ -666,6 +688,130 @@ private function processSyncRule(Rule $rule, array $data): array
...
+     * @throws \OCP\Files\InvalidPathException
+     * @throws \OCP\Files\NotFoundException
+     */
+    private function processFilePartRule(Rule $rule, array $data, Endpoint $endpoint, ?string $objectId = null): array
+    {
+        if($objectId === null) {
+            throw new Exception('Filepart rules can only be applied after the object has been created');
+        }
+
+        $config = $rule->getConfiguration()['fileparts_create'];
```
bbrands02, line 717: Can this array key fileparts_create be null? If so should it not be checked before used to prevent errors

lib/Service/EndpointService.php
```diff
@@ -666,6 +688,130 @@ private function processSyncRule(Rule $rule, array $data): array
...
+
+        $targetId = explode('/', $endpoint->getTargetId());
+
+        $registerId = $targetId[0];
+        $superSchemaId = $targetId[1];
+
+        $sizeLocation     = $config['sizeLocation'];
```
bbrands02, line 724: Same thing here from earlier comment to check if keys are set

lib/Service/EndpointService.php
```diff
@@ -666,6 +688,130 @@ private function processSyncRule(Rule $rule, array $data): array
...
+     * @throws DoesNotExistException
+     * @throws LoaderError
+     * @throws MultipleObjectsReturnedException
+     * @throws SyntaxError
+     * @throws \OCP\Files\InvalidPathException
+     * @throws \OCP\Files\NotFoundException
+     */
+    private function processFilePartUploadRule(Rule $rule, array $data, ?string $objectId = null): array
+    {
+        $config = $rule->getConfiguration()['filepart_upload'];
```
bbrands02, line 795: Same thing here as from earlier comment

lib/Service/StorageService.php
```diff
@@ -0,0 +1,231 @@
...
+     *
+     * @return bool Whether or not reconciling the file has been successful.
+     */
+    private function attemptCloseUpload (array $folderContents, File $target, int $numParts): bool
+    {
+        $contentFilenames = array_map(function ($node) {
+            return $node->getName();
+        }, $folderContents);
+
+        $folder = $folderContents[0]->getParent();
```
bbrands02, line 150: Can this ever be empty, what happens then?

lib/Service/EndpointService.php
rjzondervan: Added a check for this

lib/Service/EndpointService.php
rjzondervan: Added a check for this

lib/Service/StorageService.php
rjzondervan: The moment this function is called, folderContents contains at least _n_ files where _n_ is the number of parts of the upload, so this can never be empty.

bbrands02 (approved):

remko48 (approved):

PR #208, opened 2025-02-10
Assign tags for fetch and write rules

bbrands02:
Also has fixes in synchronizeContract for the use of $log where the log for synchronization in a earlier commit and merge got confused with the log for the contract


lib/Service/SynchronizationService.php
```diff
@@ -649,7 +659,10 @@ public function synchronizeContract(
         } else {
             $targetObject = $object;
         }
-		$log->setTarget($targetObject);
+
+        if ($contractLog) {
```
WilcoLouwerse, line 663: if $contractLog what? missing any kind of comparison here ===

lib/Service/SynchronizationService.php
```diff
@@ -687,12 +700,22 @@ public function synchronizeContract(
         }
 
 		// Create log entry for the synchronization
-		$log->setTargetResult($synchronizationContract->getTargetLastAction());
-		$log = $this->synchronizationContractLogMapper->update($log);
-		$synchronizationContract = $this->synchronizationContractMapper->update($synchronizationContract);
+        if ($contractLog) {
```
WilcoLouwerse, line 703: See previous comment

WilcoLouwerse:
Missing docu

WilcoLouwerse (approved):

PR #219, opened 2025-02-14
Config xxllnc suite + force fix

bbrands02:


lib/Service/SynchronizationService.php
```diff
@@ -1779,27 +1779,73 @@ private function processWriteFileRule(Rule $rule, array $data, string $objectId,
...
+                } else {
+                    $content = $value;
+                }
 
-			$tags = array_merge($config['tags'] ?? [], ["object:$objectId"]);
-			if ($file instanceof File === true) {
-				$this->attachTagsToFile(fileId: $file->getId(), tags: $tags);
-			}
-        } catch (Exception $exception) {
+                $content = base64_decode($content);
```
rjzondervan, line 1807: Base64 decode is already done in $objectService->addFile()

lib/Service/SynchronizationService.php
```diff
@@ -1779,27 +1779,73 @@ private function processWriteFileRule(Rule $rule, array $data, string $objectId,
...
+                    }
+
+                    $result[$key] = $file->getPath();
+                } catch (Exception $exception) {
+                }
+            }
+            $result[$key] = $file->getPath();
+            $dataDot[$config['filePath']] = $result;
+        } else { 
+            $content = base64_decode($files);
```
rjzondervan, line 1829: Base64 decode is already done in $objectService->addFile()

rjzondervan (changes requested):
Small compatibility fix

PR #227, opened 2025-02-19
Move authentication logic to rules, allowing for public endpoints

rjzondervan:


lib/Service/EndpointService.php
```diff
@@ -634,6 +638,50 @@ private function getRuleById(string $id): ?Rule
 		}
 	}
 
+    private function processAuthenticationRule (Rule $rule, array $data): array|JSONResponse
```
bbrands02, line 641: Doc blocks

lib/Service/AuthorizationService.php
```diff
@@ -198,4 +201,30 @@ public function authorize(string $authorization): void
 		$this->validatePayload($payload);
 		$this->userSession->setUser($this->userManager->get($issuer->getUserId()));
 	}
+
+    public function authorizeBasic (string $header, array $users, array $groups): void
```
bbrands02, line 205: Doc blocks

bbrands02 (approved):

PR #276, opened 2025-03-06
Added support for Endpoints to return XML Response

WilcoLouwerse:


composer.json
```diff
@@ -39,7 +39,10 @@
 		"symfony/uid": "^6.4",
 		"symfony/yaml": "^6.4",
 		"twig/twig": "^3.14",
-		"web-token/jwt-framework": "^3"
+		"web-token/jwt-framework": "^3",
+      	"ext-simplexml": "*",
+		"ext-libxml": "*",
+		"ext-dom": "*"
```
rjzondervan, line 45: Did you check these were standard enabled in Nextcloud (they are, but please check before requiring extra extensions)

lib/Http/XMLResponse.php
```diff
@@ -0,0 +1,109 @@
...
+			}
+		}
+
+		// Convert SimpleXMLElement to formatted XML with declaration
+		$dom = new DOMDocument('1.0', 'UTF-8');
+		$dom->preserveWhiteSpace = false;
+		$dom->formatOutput = true; // Enable formatting for readability
+		$dom->loadXML($xml->asXML());
+
+		return $dom->saveXML();
```
rjzondervan, line 107: You're encoding the XML manually, but isn't there a library in nextcloud to encode the XML for us? (From the viewpoint of maintainability)

composer.json
WilcoLouwerse: image
image

lib/Http/XMLResponse.php
WilcoLouwerse: image
Couldn't find anything in nextcloud to encode xml

rjzondervan (approved):

PR #281, opened 2025-03-06
Prevent deletion of sub objects each time the sync ran

bbrands02:
This fix is needed because currently after each synchronization run we check if there are invalid objects to delete, because the created sub objects are not in the array of originIds we earlier created we should not check for these to delete, and only check for contracts where the target object has the schemaId that is configured on the synchronization

The deletion of these sub objects should be handled by cascade delete or their own synchronizations


WilcoLouwerse (approved):

bbrands02:
CONNECTOR-182

PR #298, opened 2025-03-11
Keep 'before' & 'after' similair

WilcoLouwerse:


rjzondervan (approved):
Approved for now, but please state secundary changes in your PR description, as it is not clear as of why you are changing things on the mappers in this specific PR.

PR #335, opened 2025-03-20
Json editor to change the data in the frontend (editConsumer modal)

CBibop12:


src/modals/Consumer/EditConsumer.vue
```diff
@@ -106,14 +130,35 @@ export default {
...
+							? parsed.domains.join(', ')
+							: (parsed.domains || ''),
+						ips: Array.isArray(parsed.ips)
+							? parsed.ips.join(', ')
+							: (parsed.ips || ''),
+					}
+				} catch (e) {
+				}
+			},
+		},
```
SudoThijn, line 160: this does work, and is quite cool, but there are a few issues with this approach.
It would be better to keep 2 different variables and manually synchronize them.

regardless, the current feature is not what was wanted from the issue, so there is no point in keeping it / fixing it now.

src/modals/Consumer/EditConsumer.vue
SudoThijn, line 1: This is not exactly as was requested in the issue.

The issue requested that `consumerItem.authorizationConfiguration` is editable in a code-editor.
Not the entire consumerItem object.

For this, the syncing with `jsonContent` is not needed.

SudoThijn (changes requested):

src/modals/Consumer/EditConsumer.vue
```diff
@@ -82,17 +102,22 @@ export default {
...
 		return {
 			consumerItem: {
+				id: null,
 				name: '',
 				description: '',
 				domains: '',
 				ips: '',
 				authorizationType: '',
-				authorizationConfiguration: '',
+				authorizationConfiguration: [['']],
```
bbrands02, line 118: Can this be defaulted to just []

src/modals/Consumer/EditConsumer.vue
```diff
@@ -106,14 +131,22 @@ export default {
...
+	watch: {
+		authConfig(newVal) {
+			try {
+				JSON.parse(newVal)
+				this.error = false
+			} catch (e) {
+				this.error = 'Invalid JSON in authorization configuration'
+			}
+		},
+	},
```
SudoThijn, line 149: This'll also show the error notecard, which we only use for big errors from API requests.

This would be much better as a computed value.
Since computed values actually return a value, which can set to be either true or false in this case.

e.g.
```js
	computed: {
		isAuthConfigValid() {
			try {
				JSON.parse(newVal)
				return true
			} catch (e) {
				return false
			}
		},
	},
```

This you can now use to disable the submit button.
Or show a simple error message under the code-editor.

src/modals/Consumer/EditConsumer.vue
```diff
@@ -147,35 +194,50 @@ export default {
...
-
-			const newConsumer = new Consumer({
+			let parsedAuthConfig
+			try {
+				parsedAuthConfig = JSON.parse(this.authConfig)
+			} catch (e) {
+				this.error = 'Invalid JSON in authorization configuration'
+				this.loading = false
+				return
+			}
```
SudoThijn, line 222: The moment you get here, the auth config should be valid anyway, so there is no reason to check again if it's valid.

src/modals/Consumer/EditConsumer.vue
```diff
@@ -147,35 +194,50 @@ export default {
...
 				name: '',
 				description: '',
 				domains: '',
 				ips: '',
 				authorizationType: '',
-				authorizationConfiguration: '',
+				authorizationConfiguration: [['']],
 			}
 			this.authorizationTypeOptions.value = { label: 'none' }
+			this.authConfig = JSON.stringify([['']], null, 2)
```
SudoThijn, line 206: Also make sure to properly update the closeModal function in accordance with my previous points.

SudoThijn (changes requested):

PR #344, opened 2025-03-21
Make extra data for mapping backwards compatible

bbrands02:


lib/Service/EndpointService.php
```diff
@@ -102,20 +111,48 @@ public function handleRequest(Endpoint $endpoint, IRequest $request, string $pat
...
-					$request->getHeader('Content-Type')
-				),
-			];
+                'parameters' => $incomingParams,
+                'headers' => $incomingHeaders,
+                'path' => $path,
+                'method' => $incomingMethod,
+                'body' => array_merge([
+                    '_utility' => [
+                        'currentDate' => $currentDate
```
WilcoLouwerse, line 149: 2x currentDate?
image

rjzondervan (approved):

lib/Service/EndpointService.php
bbrands02: Yes the keys in the root are for other rules and the keys in body with _ are for mapping

PR #353, opened 2025-03-24
Some more fixes / changes for correct XMLResponse

WilcoLouwerse:


lib/Http/XMLResponse.php
```diff
@@ -136,8 +136,16 @@ public function arrayToXml(array $data, ?string $rootTag = null): string
 		// Build XML structure
 		$this->buildXmlElement($dom, $root, $data);
 		
-		// Convert DOM to string
-		return $dom->saveXML() ?: '';
+		// Get XML output
+		$xmlOutput = $dom->saveXML() ?: '';
+		
+		// Directly replace decimal CR entities with hexadecimal
+		$xmlOutput = str_replace('&#13;', '&#xD;', $xmlOutput);
```
bbrands02, line 143: Are these html encoded characters? Should these not be done with mapping? Feels risky to replace all perhaps

tests/Http/XMLResponseTest.php
```diff
@@ -85,7 +85,16 @@ public function arrayToXml(array $data, ?string $rootTag = null): string {
         
         $this->buildXmlElement($dom, $root, $data);
         
-        return $dom->saveXML() ?: '';
+        // Get XML output
+        $xmlOutput = $dom->saveXML() ?: '';
+        
+        // Directly replace decimal CR entities with hexadecimal
+        $xmlOutput = str_replace('&#13;', '&#xD;', $xmlOutput);
```
bbrands02, line 92: Same hting

tests/Http/XMLResponseTest.php
```diff
@@ -469,6 +509,43 @@ private function testArchiMateOpenGroupModelXML(): void
...
+     * Tests:
+     * - lib/Http/XMLResponse.php::arrayToXml (empty tag formatting)
+     * 
+     * @return void
+     */
+    private function testEmptyTagFormatting(): void
+    {
+        echo "- Running testEmptyTagFormatting: ";
+        
+        $data = [
```
bbrands02, line 524: Maybe make this a class const

lib/Http/XMLResponse.php
rjzondervan: Actually, these are equivalent characters, &#13 and &#xD both resolve to the same character, only &#13 is the decimal value, and &#xD is the hexadecimal value.
The best way would be if we were able to choose the type of html encoding (decimal or hexadecimal), so for now I'd say this is fine.

rjzondervan (approved):

bbrands02 (approved):

PR #364, opened 2025-03-27
Only add Content-Disposition header to XMLResponse if url ends with .xml

WilcoLouwerse:


lib/Http/XMLResponse.php
```diff
@@ -49,9 +51,13 @@ public function __construct($data = [], int $status = 200, array $headers = [])
 			$this->addHeader($name, $value);
 		}
 		
-		// Set content type and disposition headers
+		// Set content type header
 		$this->addHeader('Content-Type', 'application/xml; charset=utf-8');
-		$this->addHeader('Content-Disposition', 'attachment; filename="export.xml"');
+		
+		// Only add Content-Disposition header if path ends with .xml
+		if ($path !== null && str_ends_with($path, '.xml')) {
```
bbrands02, line 58: === true

rjzondervan (approved):

PR #401, opened 2025-03-31
Fix logging endpoint of synchronizations, fix setting last synced

rjzondervan:


WilcoLouwerse (approved):
Can be merged with approval from PO (in slack is approval is also ok)

bbrands02 (approved):

PR #417, opened 2025-04-08
renderEntity each object

bbrands02:


WilcoLouwerse (approved):
You could also add support for 'fields' here :)

PR #435, opened 2025-04-10
Api logo is color-coded

CBibop12:


src/views/Endpoint/EndpointsList.vue
```diff
@@ -124,6 +125,24 @@ export default {
...
+				return '#87547a'
+			case 'PATCH':
+				return '#a95d2e'
+			case 'DELETE':
+				return '#b13f3a'
+			default:
+				return '#000'
+			}
+		},
+	},
```
SudoThijn, line 145: Good implementation.

But these colors currently only support light mode, as the contrast quality on dark mode is not good.
Please use the `getTheme()` function to change the color depending on the theme.
Dark mode should use these colors:
- get: #5c8d4a
- post: #5d82c0
- put: #a46f96
- delete: #d25c53
- patch: #bc6d3d

While light mode can remain the same.

You can find good contrasts by using this site https://www.learnui.design/tools/accessible-color-generator.html
It allows you to find contrast with either AA or AAA guidelines, AAA is better for accessibility, but AA just looks better and is still acceptable.

SudoThijn (changes requested):

PR #436, opened 2025-04-11
Configuration Service and Mapper cleanup

rubenvdlinde:


WilcoLouwerse:
@rubenvdlinde i think we can review and merge this 🤔

PR #454, opened 2025-04-23
Added more config for the update and delete external sync

bbrands02:


rjzondervan (approved):
Seems okay, but check about throwEvent with @rubenvdlinde

WilcoLouwerse (approved):

PR #457, opened 2025-04-28
Allow setting the source ad hock for a synchronisation in the synchronysation service and allow the passing trought he controller

rubenvdlinde:
Allow setting the source (location) ad hoc for a synchronization in the SynchronizationService, and allow passing a source (location) and additional object data (e.g., documentation URL) in the SynchronizationController.

Small note:
This still requires the current Software Catalog frontend to pass the source (location) as the source parameter and the documentation base URL as the data.documentation_url parameter.

Also, this will now (by default) only work on a force, since the entire validation and contracting setup still needs to be refactored to handle multiple sources for a synchronization.


WilcoLouwerse:
> https://github.com/ConductionNL/openconnector/pull/456

PR #465, opened 2025-05-01
Fix for download

WilcoLouwerse:


lib/Http/XMLResponse.php
```diff
@@ -233,6 +234,17 @@ private function createChildElement(DOMDocument $dom, DOMElement $parentElement,
 		if (is_array($data) === true) {
 			$this->buildXmlElement($dom, $childElement, $data);
 		} else {
+			// Handle objects that might not be convertible to string directly
+			if (is_object($data) === true) {
+				// For QueryBuilder objects or objects without __toString(), create a placeholder
+				if ($data instanceof \OCP\DB\QueryBuilder\IQueryBuilder ||
```
bbrands02, line 240: Import IQueryBuilder?

lib/Http/XMLResponse.php
```diff
@@ -233,6 +234,17 @@ private function createChildElement(DOMDocument $dom, DOMElement $parentElement,
 		if (is_array($data) === true) {
 			$this->buildXmlElement($dom, $childElement, $data);
 		} else {
+			// Handle objects that might not be convertible to string directly
+			if (is_object($data) === true) {
+				// For QueryBuilder objects or objects without __toString(), create a placeholder
+				if ($data instanceof \OCP\DB\QueryBuilder\IQueryBuilder || 
+					!method_exists($data, '__toString')) {
```
bbrands02, line 241: === false

lib/Service/RuleService.php
```diff
@@ -154,7 +176,41 @@ private function processSoftwareCatalogusRule(Rule $rule, array $data): array
...
+        //                 'propertyDefinitionRef' => 'propid-39', // URL
+        //                 'value' => '',
+        //             ],
+        //         ],
+        //     ];
+        // }
+
+        // foreach ($voorzieningAanbod as $voorzieningAanbod) {
+        //     $voorzieningAanbod = $voorzieningAanbod->jsonSerialize();
         // }
```
rjzondervan, line 214: Is there a reason for this long commented piece of code?

WilcoLouwerse: This code has not been tested enough, it is out of scope for now, but we will need it later on. For now commented out because we quickly want to merge a fix that will make sure the Exports/downloads are valid to be imported into Archi

bbrands02 (approved):

rjzondervan (approved):

PR #498, opened 2025-05-27
Speeding up the synchronisation 

rubenvdlinde:
From 60.000 tot 6.000 ms for 20 cases including files

### DIFF
--- a/lib/Service/SynchronizationService.php
+++ b/lib/Service/SynchronizationService.php
@@ -1985,7 +1985,11 @@ class SynchronizationService
 
 		// Write file with OpenRegister FileService.
 		$fileService = $this->containerInterface->get('OCA\OpenRegister\Service\FileService');
-		$file = $fileService->addFile(objectEntity: $objectEntity, fileName: $filename, content: $response['body'], share: isset($config['autoShare']) ? $config['autoShare'] : false, tags: $tags);
+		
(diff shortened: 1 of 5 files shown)


lib/Service/SynchronizationService.php
```diff
@@ -2098,89 +2379,225 @@ private function getArrayType(mixed $array): string
...
+                return 'file://fetching-async';
+                
+            case "Multidimensional array":
+                return array_fill(0, count($endpoint), 'file://fetching-async');
+                
+            case "Indexed array":
+                return array_fill(0, count($endpoint), 'file://fetching-async');
+                
+            default:
+                return 'file://fetching-async';
```
rjzondervan, line 2597: switches accept multiple cases for the same execution, so this could be condensed into:

```php
        switch ($this->getArrayType($endpoint)) {
            case "Multidimensional array":
            case "Indexed array":
                return array_fill(0, count($endpoint), 'file://fetching-async');

            case 'Not array':
            case 'Associative array':
            default:
                return 'file://fetching-async';
       }
```

rjzondervan (approved):
Seems in order, small code style advisory.

lib/Service/SynchronizationService.php
rubenvdlinde: Fair, there are more code styling issues here. But it was i needed nightly so pick this up in the refactor

PR #535, opened 2025-07-07
Fix rewriting filters

rjzondervan:
Rewrite them in a way that the form in the database does not matter.
Also, fix importing jobs

### DIFF
--- a/lib/Service/EndpointService.php
+++ b/lib/Service/EndpointService.php
@@ -495,6 +495,9 @@ class EndpointService
 
         $rewriteParameters = array_intersect(array_keys($parameters), array_keys($schema->getProperties()));
 
+//        var_dump($rewriteParameters);
+
+
         foreach($rewriteParameters as $rewriteParameter) {
             if (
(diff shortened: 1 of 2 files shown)


lib/Service/EndpointService.php
```diff
@@ -495,6 +495,9 @@ private function rewriteExternalReferences(array $parameters, \OCA\OpenRegister\
 
         $rewriteParameters = array_intersect(array_keys($parameters), array_keys($schema->getProperties()));
 
+//        var_dump($rewriteParameters);
```
bbrands02, line 498: commented dump

lib/Service/EndpointService.php
```diff
@@ -503,6 +506,13 @@ private function rewriteExternalReferences(array $parameters, \OCA\OpenRegister\
                 continue;
             }
 
+//            var_dump($schema->getProperties()[$rewriteParameter]);
```
bbrands02, line 509: commented dump and code

bbrands02 (approved):

PR #599, opened 2025-09-30
Fixes for event listener and correct 404 error

rjzondervan:


lib/Service/EndpointService.php
```diff
@@ -1276,6 +1282,7 @@ private function processMapping(Rule $rule, Mapping $mapping, array $data): arra
 
             return $data;
         }
+//        var_dump($data['body']);
```
bbrands02, line 1285: dump

PR #612, opened 2025-10-22
openconnector deeplinking

SudoThijn:


rjzondervan (approved):

SudoThijn:
In my previous commit I converted the Legacy ESLint config to a Flat Config.
I also added a new helper package [`@eslint/config-helpers`](https://www.npmjs.com/package/@eslint/config-helpers) to ensure compatibility with older versions of ESLint.

This has to be done due to the recent change in the Nextcloud app itself, as they have also converted to Flat Config, their config is overwriting ours.
ESLint prefers Flat Config over Legacy. So as to not have an unrelated config take priority, we have to convert to the new Flat Config ourselves

PR #628, opened 2025-11-27
Parallelise replacing internal calls and set a default limit on calls from OpenConnector

rjzondervan:


lib/Service/CallService.php
```diff
@@ -56,17 +65,45 @@ public function __construct(
...
+     *
+     * @param int[] $retentions The list of retentions in milliseconds to find the maximum duration for.
+     * @return \DateTime|null The calculated expiry
+     * @throws \DateMalformedStringException
+     *
+     * @TODO: At a later point in time this should be changed to using the most specific source for expiration
+     */
+    private function calculateExpires(...$retentions): ?\DateTime
+    {
+        if (in_array(0, $retentions, true)) {
```
WilcoLouwerse: === true

lib/Service/JobService.php
```diff
@@ -102,21 +74,41 @@ class JobService
...
+        IAppConfig $appConfig,
     ) {
-        $this->jobList = $jobList;
-        $this->jobMapper = $jobMapper;
-        $this->connection = $connection;
-        $this->jobLogMapper = $jobLogMapper;
-        $this->containerInterface = $containerInterface;
-        $this->userSession = $userSession;
-        $this->userManager = $userManager;
+        if($appConfig->hasKey(app: 'openconnector', key: 'retention')) {
```
WilcoLouwerse: if (

lib/Service/JobService.php
```diff
@@ -102,21 +74,41 @@ class JobService
...
+     *
+     * @param int[] $retentions The list of retentions in milliseconds to find the maximum duration for.
+     * @return \DateTime|null The calculated expiry
+     * @throws \DateMalformedStringException
+     *
+     * @TODO: At a later point in time this should be changed to using the most specific source for expiration
+     */
+    private function calculateExpires(...$retentions): ?\DateTime
+    {
+        if (in_array(0, $retentions, true)) {
```
WilcoLouwerse: === true

PR #713, opened 2026-04-14
Enable twig for endpoint

bbrands02:


rjzondervan (approved):

WilcoLouwerse (commented):
## PR Review

**SSTI Risk** — The endpoint string is passed directly to Twig rendering. If `renderTemplateString()` doesn't use a sandboxed environment, any user who can configure `sourceConfig['endpoint']` could execute arbitrary code. Verify sandboxing before merging.

**Rate-limit bypass** — The reordering moves `checkRateLimit()` to *after* template rendering. This means rate-limited requests still do Twig processing before being rejected. Keep the rate-limit check at the top for fail-fast behavior.

WilcoLouwerse (approved):

PR #714, opened 2026-04-14
Prevent fetched object double wrapping

bbrands02:
prevents 
[[object]]

when dealing with a single fetched object

### DIFF
--- a/lib/Service/SynchronizationService.php
+++ b/lib/Service/SynchronizationService.php
@@ -628,10 +628,12 @@ class SynchronizationService
             $stageStartTime = microtime(true);
             $result['objects']['found'] = count($objectList);
 
-            if ($sourceConfig['resultsPosition'] === '_object') {
-                $objectList = [$objectList];
-                $result['objects']['found'] = count($objectList);
-            }
+	            if ($sourceConfig['resultsPosition'] === '_object') {
+	                if (array_is_list($objectList) === false) {
+	                    $objectList = [$objectList];
+	                }
+	                $result['objects']['found'] = count($objectList);
+	            }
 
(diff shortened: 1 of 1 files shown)


WilcoLouwerse:
Nice fix for the double-wrapping issue — `array_is_list()` is the right guard here.

Two things I noticed:

**1. Indentation mismatch**
The new blocks at both locations use tabs where the surrounding code uses spaces. Visible in the diff — the `if` block around line 632 and the replacement around line 1949 are tab-indented inside space-indented context. Worth fixing before merge.

**2. Wrapping logic split across two methods**
The call chain is `synchronizeExternToIntern` → `getAllObjectsFromSource` → `getAllObjectsFromApi`. With this change, the `_object` wrapping is skipped in `getAllObjectsFromApi` and handled by the caller instead. That works, but it means anyone calling `getAllObjectsFromApi` directly won't get the normalization for `_object` results.

Would it be cleaner to keep all the normalization in `getAllObjectsFromApi` — i.e., do the `array_is_list()` check there for *all* `resultsPosition` values, including `_object` — so it always returns a proper list regardless of who calls it?

WilcoLouwerse:
Note: all 6 failing CI jobs are unrelated to this PR.

- **PHPCS, PHPStan, PHPMD, Psalm, PHPUnit** — runner has PHP 8.1.34 but composer requires >= 8.3. Looks like a missing `setup-php` step in the workflow.
- **Frontend Quality** — transient `ECONNRESET` during `npm install`.

These fail on `main` too.

WilcoLouwerse (approved):

rjzondervan (approved):

PR #718, opened 2026-04-20
refactored dashboard page

SudoThijn:


l10n/en.json
```diff
@@ -77,7 +77,9 @@
         "Could not update contract: %s": "Could not update contract: %s",
         "Created": "Created",
         "Critical": "Critical",
+        "Dashboard": "Dashboard",
         "Date Range": "Date Range",
+        "Date range": "Date range",
```
remko48: This looks like double translations is there a specific instance where we cannot use on or the other?

l10n/nl.json
```diff
@@ -77,7 +77,9 @@
         "Could not update contract: %s": "Kan contract niet bijwerken: %s",
         "Created": "Aangemaakt",
         "Critical": "Kritiek",
+        "Dashboard": "Dashboard",
         "Date Range": "Datumbereik",
+        "Date range": "Datumbereik",
```
remko48, line 81: This looks like double translations is there a specific instance where we cannot use on or the other?

l10n/en.json
```diff
@@ -156,6 +158,7 @@
         "First": "First",
         "Fout": "Fout",
         "From": "From",
+        "From:": "From:",
```
remko48: I would just use the `"From"` translation and add the `:` at the end of the string in code

l10n/en.json
```diff
@@ -350,13 +359,16 @@
...
+        "Synchronization executions (daily)": "Synchronization executions (daily)",
+        "Synchronization executions (hourly)": "Synchronization executions (hourly)",
         "Synchronizations": "Synchronizations",
         "Target": "Target",
         "Target ID": "Target ID",
         "Target Type": "Target Type",
         "The specified schema could not be found.": "The specified schema could not be found.",
         "Time Sensitive": "Time Sensitive",
         "To": "To",
+        "To:": "To:",
```
remko48: Same as the comment about `"From:"`

l10n/nl.json
```diff
@@ -156,6 +158,7 @@
         "First": "Eerste",
         "Fout": "Fout",
         "From": "Van",
+        "From:": "Van:",
```
remko48: I would just use the `"From"` translation and add the `:` at the end of the string in code

l10n/nl.json
```diff
@@ -350,13 +359,16 @@
...
+        "Synchronization executions (daily)": "Synchronisatie-uitvoeringen (dagelijks)",
+        "Synchronization executions (hourly)": "Synchronisatie-uitvoeringen (per uur)",
         "Synchronizations": "Synchronisaties",
         "Target": "Doel",
         "Target ID": "Doel-ID",
         "Target Type": "Doeltype",
         "The specified schema could not be found.": "Het opgegeven schema kon niet worden gevonden.",
         "Time Sensitive": "Tijdgevoelig",
         "To": "Tot",
+        "To:": "Tot:",
```
remko48: Same as the comment about "From:"

src/views/dashboard/DashboardIndex.vue
```diff
@@ -1,131 +1,178 @@
...
+			:title="t('openconnector', 'Dashboard')"
+			:widgets="widgetDefs"
+			:layout="dashboardLayout"
+			:loading="isLoading && !hasData"
+			@layout-change="onLayoutChange">
+			<template #widget-stat-sources>
+				<CnStatsBlock
+					title="Sources"
+					:count="stats.sources || 0"
+					count-label="sources"
```
remko48: I would make sure that we only show the plural of source if there is more than one. 
Now we have `1 sources` this should be `1 source` 
```suggestion
					:count-label="stats.sources === 1 ? 'source' : 'sources'"
```

Also apply this to the other StatsBlocks

remko48 (changes requested):
Looks good, just some small things like duplicates and eslint

l10n/en.json
SudoThijn: The difference is the "r", one is capital and one is not.
Though it's true that this is useless, i'll fix it and just have "Date range"

remko48 (approved):

PR #724, opened 2026-04-23
Fix forward compatibility of SyncService

rjzondervan:


bbrands02 (dismissed):

bbrands02 (approved):

lib/Service/SynchronizationService.php
```diff
@@ -2989,7 +2989,7 @@ private function fetchFile(Source $source, string $endpoint, array $config, stri
 
 		try {
 			$objectService = $this->containerInterface->get('OCA\OpenRegister\Service\ObjectService');
-			$objectEntity = $objectService->findByUuid(uuid: $objectId);
+			$objectEntity = $objectService->find(id: $objectId);
```
WilcoLouwerse, line 2992: Using `find()` as a drop-in for `findByUuid` changes behavior: it creates an audit-trail entry on every call, runs RBAC `checkPermission()`, renders the entity, and returns `?ObjectEntity` (nullable) instead of throwing on miss. For these internal sync operations that's unwanted overhead, and in background-job context the RBAC check may fail where `findByUuid` previously succeeded.

`findSilent(id: $objectId)` has the same signature but skips audit trails and rendering, and returns non-nullable `ObjectEntity`. Its OREG docblock explicitly says it's "used internally by other operations… without logging the read action" — that looks like the more faithful replacement across all four call sites here.

lib/Service/SynchronizationService.php
```diff
@@ -3498,7 +3498,7 @@ private function processWriteFileRule(Rule $rule, array $data, string $objectId,
 
         // Get the object entity and file service
         $objectService = $this->containerInterface->get('OCA\OpenRegister\Service\ObjectService');
-        $objectEntity = $objectService->findByUuid(uuid: $objectId);
+        $objectEntity = $objectService->find(id: $objectId);
```
WilcoLouwerse, line 3501: This call site has no surrounding try/catch, so with `find()` the new null-return path (and any new RBAC/permission exception) could surface differently than the old `findByUuid` behavior. Worth either handling null explicitly or switching to `findSilent()`.

.github/workflows/pull-request-from-branch-check.yaml
```diff
@@ -2,31 +2,8 @@ name: Branch Protection
...
-            if [[ "$SOURCE" != "development" ]] && ! [[ "$SOURCE" =~ ^hotfix ]]; then
-              echo "Error: Pull requests to beta must come from 'development' or a branch starting with 'hotfix'"
-              echo "Source branch: $SOURCE"
-              exit 1
-            fi
-          fi
-
-          echo "Branch check passed: $SOURCE -> $TARGET"
+  branch-protection:
+    uses: ConductionNL/.github/.github/workflows/branch-protection.yml@main
```
WilcoLouwerse, line 9: Two notes on this workflow change:

1. It's unrelated to the forward-compat fix in the PR title — might be cleaner as a separate PR.
2. `@main` pins to a moving branch, so branch-protection behavior can change silently. Pinning to a SHA or tag is safer for supply-chain stability.

WilcoLouwerse (approved):

lib/Service/SynchronizationService.php
rjzondervan: It does not change behaviour, findByUuid was always an alias for find(), so this is just cutting out an alias.

lib/Service/SynchronizationService.php
rjzondervan: As said before, findByUuid was an alias for find, so nothing changed, except the called name.

.github/workflows/pull-request-from-branch-check.yaml
rjzondervan: Change was necessary to be able to merge this PR.

lib/Service/SynchronizationService.php
WilcoLouwerse: Good catch — confirmed in OREG git history that `findByUuid` was literally `return $this->find($uuid);`. Pure alias, no regression. `findSilent()` would still skip audit-trail/render for internal sync reads, but that's a separate optimization — not something to block this hotfix on.

lib/Service/SynchronizationService.php
WilcoLouwerse: Agreed — since `findByUuid` was an alias for `find()`, the null-return path was already there. No regression.

.github/workflows/pull-request-from-branch-check.yaml
WilcoLouwerse: Fair enough — taking at face value given the merge-block requirement.

PR #725, opened 2026-04-24
Updated workflow to only trigger on pr create and on push

remko48:


bbrands02 (approved):

.github/workflows/code-quality.yml
```diff
@@ -1,70 +1,30 @@
 name: Code Quality
 
 on:
+  push:
+    branches: [main, master, beta, development, feature/**, bugfix/**, hotfix/**]
   pull_request:
-    branches: [main, master, development]
+    types: [opened, reopened]
```
WilcoLouwerse, line 7: 🟡 **Concern — CI gap on subsequent commits for non-pattern-named PR branches**

Removing `synchronize` from `pull_request.types` means subsequent pushes to an open PR only get CI via the `push` event. The push filter covers `feature/**`, `bugfix/**`, `hotfix/**`, `master`, `main`, `beta`, and `development`. PRs opened from branches like `patch/1.2`, `release/v2`, or `dependabot/npm_and_yarn/…` will receive CI once on `pull_request: opened` but not on any follow-up commits.

**Impact:** A developer pushing a fix after review feedback gets no quality signal before merge on out-of-pattern branches.

**Suggested fix:** Restore `synchronize` alongside `opened` and `reopened`:
```yaml
pull_request:
  types: [opened, reopened, synchronize]
  branches: [main, master, beta, development]
```
The existing `if` guard already prevents the new-branch-push + PR-open double-run, so adding `synchronize` back does not reintroduce that problem.

.github/workflows/code-quality.yml
```diff
@@ -1,70 +1,30 @@
...
-        run: npm run lint
-
-      - name: Stylelint
-        run: npm run stylelint
+  quality:
+    if: github.event_name != 'push' || github.event.created != true
+    uses: ConductionNL/.github/.github/workflows/quality.yml@main
+    with:
+      app-name: openconnector
+      php-version: "8.3"
```
WilcoLouwerse, line 21: 🟡 **Concern — PHP version upgraded from 8.1 → 8.3 without explicit rationale**

The old inline workflow pinned `php-version: '8.1'`. This PR upgrades to `8.3` and adds `8.4` to the PHPUnit matrix. PHP 8.1 is EOL so the upgrade is likely correct, but the change is silent — no PR description note, no linked issue.

**Impact:** If `composer.json` doesn't allow `>=8.3`, or if any code path relies on 8.1-era behaviour, this surfaces as unexpected red CI rather than a deliberate reviewed decision.

**Suggested fix:** Confirm the `composer.json` `require.php` constraint covers `>=8.3` and that the app has been validated on PHP 8.3 before merging. A one-line note in the PR description would help future readers understand the upgrade was intentional.

.github/workflows/code-quality.yml
```diff
@@ -1,70 +1,30 @@
...
-      - name: Install dependencies
-        run: npm ci
-
-      - name: ESLint
-        run: npm run lint
-
-      - name: Stylelint
-        run: npm run stylelint
+  quality:
+    if: github.event_name != 'push' || github.event.created != true
```
WilcoLouwerse, line 17: 🟢 **Minor — New `license-check` gate silently enabled via reusable workflow default**

The reusable workflow defaults `enable-license-check: true`. The old inline workflow had no equivalent gate. Since this PR does not set `enable-license-check`, the gate is implicitly activated.

This is probably intentional as part of adopting the shared standard, but it is invisible in the diff. Consider adding `enable-license-check: true` explicitly (matching the `enable-phpmetrics: true` pattern already present) to make the intent visible, and verify that existing dependencies pass the gate before merging.

WilcoLouwerse (changes requested):
2 concerns require attention before merge — the removal of `synchronize` creates a CI coverage gap for out-of-pattern branches, and the silent PHP 8.1 → 8.3 upgrade needs explicit confirmation that composer.json allows it. All old check gates are preserved by the reusable workflow.

WilcoLouwerse (approved):

PR #726, opened 2026-04-24
fix: resolve cross-entity slug references during configuration import

bbrands02:
## Summary

- **ConfigurationService**: Added `registerImportedEntityMapping()` to update in-memory slug/ID maps after each entity is imported, so entities created earlier in the same run are immediately available as references for later ones. Added `withComponentSlug()` to normalize missing slug fields from the component key. Added `reconcileImportedReferences()` for a targeted second pass over endpoints and synchronizations, which commonly depend on entities imported just before them.
- **RuleHandler**: Added `synchronization` to the entity-type lists used during slug↔ID conversion, so synchronization references in rule configs are correctly serialized and deserialized.
- **EndpointHandler**: Replaced the simple slug-only rule resolver with `resolveRuleIdentifier()`, which falls back to numeric ID, DB name lookup, and DB reference lookup — preventing silent rule drops when the source export format varies. Also injects `RuleMapper` to support the DB lookups.
- **EndpointMapper**: Initialize `$updateRegex = false` before the update loop to avoid an undefined-variable notice.
- **routes.php**: Narrow the SPA catch-all regex to exclude `/api/…` paths so JSON API endpoints resolve to their resource controllers instead of being swallowed by the Vue app route.

## Test plan

- [ ] Import a configuration bundle that includes synchronizations referenced by rules; verify the synchronization IDs resolve correctly after import
- [ ] Import a bundle where endpoints reference rules by name or reference string (not just slug); verify no rules are silently dropped
- [ ] Import a bundle with cross-entity dependencies (e.g. endpoint referencing a synchronization imported in the same run); verify the second reconciliation pass resolves the reference without a double import
- [ ] Confirm `/api/…` REST calls are no longer captured by the SPA catch-all and return JSON as expected
- [ ] Run existing import-related test suites and confirm no regressions

### DIFF
--- a/appinfo/routes.php
+++ b/appinfo/routes.php
@@ -124,7 +124,8 @@ return [
 		['name' => 'ui#cloudEventsEventsId', 'url' => '/cloud-events/events/{id}', 'verb' => 'GET'],
 		['name' => 'ui#cloudEventsLogs', 'url' => '/cloud-events/logs', 'verb' => 'GET'],
 		['name' => 'ui#import', 'url' => '/import', 'verb' => 'GET'],
-		// SPA catch-all — serves the Vue app for any frontend route (history mode routing)
-		['name' => 'dashboard#page', 'url' => '/{path}', 'verb' => 'GET', 'requirements' => ['path' => '.+'], 'defaults' => ['path' => '']],
+		// SPA catch-all — serves the Vue app for frontend routes only.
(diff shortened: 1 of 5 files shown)


bbrands02:
Closing — scope was too broad. Recreating with only ConfigurationService and RuleHandler changes.

PR #727, opened 2026-04-24
fix: resolve cross-entity slug references during configuration import

bbrands02:
## Summary

- **ConfigurationService**: Added `registerImportedEntityMapping()` to update in-memory slug/ID maps after each entity is imported, so entities created earlier in the same run are immediately available as references for later ones. Added `withComponentSlug()` to normalize missing slug fields from the component key. Added `reconcileImportedReferences()` for a targeted second pass over endpoints and synchronizations, which commonly depend on entities imported just before them.
- **RuleHandler**: Added `synchronization` to the entity-type lists used during slug↔ID conversion, so synchronization references in rule configs are correctly serialized and deserialized.

## Test plan

- [x] Import a configuration bundle that includes synchronizations referenced by rules; verify the synchronization IDs resolve correctly after import
- [x] Import a bundle with cross-entity dependencies (e.g. endpoint referencing a synchronization imported in the same run); verify the second reconciliation pass resolves the reference without a double import
- [x] Run existing import-related test suites and confirm no regressions


lib/Service/ConfigurationService.php
```diff
@@ -753,51 +841,65 @@ public function importConfiguration(array $oas): array
...
             }
         }
 
         // 3. Import rules (depends on sources).
         if (isset($components['rules'])) {
             foreach ($components['rules'] as $ruleSlug => $ruleData) {
+                $ruleData = $this->withComponentSlug($ruleData, $ruleSlug);
                 $rule = $this->handlers['rule']->import($ruleData, $this->mappings);
                 $result['rules'][$ruleSlug] = $rule;
+                $this->registerImportedEntityMapping('rule', $rule);
```
WilcoLouwerse, line 867: 🟡 **Concern — `registerImportedEntityMapping` calls for `'rule'` and `'job'` are dead code**

`$this->mappings` only contains buckets initialised by `resetMappings()` (source, mapping, endpoint, synchronization, and similar mapper-backed types). There is no `'rule'` or `'job'` bucket, so both calls hit the early-return guard (`isset($this->mappings[$entityType]) === false`) and silently no-op. The guard prevents corruption, but the calls imply rule and job slugs participate in cross-entity slug resolution — which they currently do not.

Same issue applies to the `registerImportedEntityMapping('job', $job)` call a few lines below.

Either remove both dead calls, or — if rule/job slugs are intended to be resolvable in the future — add the corresponding buckets to `resetMappings()`.

lib/Service/ConfigurationService.php
```diff
@@ -218,6 +219,93 @@ private function resetMappings(): void
...
+            foreach ($components['rules'] as $ruleSlug => $ruleData) {
+                $ruleData = $this->withComponentSlug($ruleData, $ruleSlug);
+                $result['rules'][$ruleSlug] = $this->handlers['rule']->import($ruleData, $this->mappings);
+            }
+        }
+
+        if (isset($components['endpoints']) && is_array($components['endpoints'])) {
+            foreach ($components['endpoints'] as $endpointSlug => $endpointData) {
+                $endpointData = $this->withComponentSlug($endpointData, $endpointSlug);
+                $result['endpoints'][$endpointSlug] = $this->handlers['endpoint']->import($endpointData, $this->mappings);
```
WilcoLouwerse, line 304: 🟡 **Concern — `reconcileImportedReferences` doesn't re-register updated endpoint mappings after the second-pass import**

After re-importing endpoints here, `registerImportedEntityMapping` is not called on the updated entity. If the upsert produces a freshly-assigned ID (e.g. the first-pass import failed to fully resolve the entity and the second pass creates it), `$this->mappings['endpoint']` will be stale.

No current downstream step reads from the endpoint mapping after the reconcile pass runs, so there is no observable bug today. However, if `reconcileImportedReferences` is extended to include additional entity types that reference endpoints, the stale mapping will cause hard-to-trace failures.

Consider calling `$this->registerImportedEntityMapping('endpoint', $result['endpoints'][$endpointSlug])` after this line.

WilcoLouwerse (approved):
The cross-entity slug reconciliation logic is sound and the ordering rationale is well-documented; two non-blocking concerns around dead registerImportedEntityMapping calls and a stale endpoint mapping after the second pass are worth addressing but do not block merge.

PR #733, opened 2026-04-28
Fix sync cleanup scope check + sync development with main

### LINKED ISSUE #732: [OpenSpec] [openconnector] fix-sync-cleanup-scope-check
**OpenSpec change**: `fix-sync-cleanup-scope-check`

Synchronizations no longer delete OpenRegister objects when they disappear from the source. The cleanup query in `SynchronizationContractMapper::findAllBySynchronizationAndSchema` joins against the legacy `openregister_objects` table, which OpenRegister has stopped writing to since the move to per-register/schema "magic tables" (`oc_or_{registerId}_{schemaId}`). The JOIN therefore returns zero rows on every sync, the cleanup diff is always empty, and orphan objects accumulate indefinitely.

A naive fix that simply drops the JOIN is unsafe: `ObjectService::deleteObject($uuid)` resolves UUIDs across all magic tables, so any contract whose `target_id` points outside the synchronization's current `register/schema` scope (e.g. after the sync's target was changed mid-life, or from legacy data) would result in cross-schema object deletion.

**Solution**: replace the JOIN with a `synchronization_id`-only candidate query, then verify each candidate is in scope via `ObjectService::find($targetId, register: $registerId, schema: $schemaId)` before calling `deleteObject`. Stop swallowing `DoesNotExistException` silently in the cleanup loop.

See full artifacts in `openspec/changes/fix-sync-cleanup-scope-check/` (`proposal.md`, `design.md`, `specs/synchronization-cleanup/spec.md`, `tasks.md`).

## Specs

- **`synchronization-cleanup`** — Behavior of the synchronization cleanup pass: which contracts are considered for deletion, what scope guards apply, and how out-of-scope or missing target objects are handled. Four requirements covering happy-path delete, scope-respect, public-API-only, and partial-failure logging.

### 1. Mapper rewrite

- [x] **1.1** In `lib/Db/SynchronizationContractMapper.php`, rename `findAllBySynchronizationAndSchema` to `findAllBySynchronization`. Drop the `$schemaId` parameter from the signature and update the docblock.
- [x] **1.2** Replace the method body with a single-table query: select all columns from `openconnector_synchronization_contracts` filtered only by `synchronization_id`. Remove the alias `'c'`, the `INNER JOIN openregister_objects`, and the `o.schema = :schemaId` predicate.
- [x] **1.3** Verify the method still returns an array of `SynchronizationContract` entities (or `[]` on exception) so the call site does not need changes beyond the rename.

### 2. Cleanup pass — scope-checked deletion

- [x] **2.1** In `lib/Service/SynchronizationService.php::deleteInvalidObjects`, update the call site (currently around line 1155) from `findAllBySynchronizationAndSchema(syncId, schemaId)` to `findAllBySynchronization(syncId)`.
- [x] **2.2** Inside the `case 'register/schema':` branch, after computing `$targetIdsToDelete`, before the `foreach ($targetIdsToDelete ...)` loop, resolve the `ObjectService` instance the same way the rest of `SynchronizationService` does (via the container).
- [x] **2.3** In the `foreach`, before calling `$this->updateTarget(...)` for each candidate, call `$objectService->find($targetIdToDelete, register: $registerId, schema: $schemaId)`. Resolved: passes `_rbac: false, _multitenancy: false` to match the prior SQL JOIN's lack of authorization filtering.
- [x] **2.4** Replace the swallowed `catch (DoesNotExistException $exception) { // @todo log }` block with a `LoggerInterface::warning(...)` call that includes synchronization id, target id, and exception message. (`LoggerInterface` was already injected as `$this->logger`.)

### 3. Tests

- [x] **3.1** Unit test for `SynchronizationContractMapper::findAllBySynchronization` (`tests/Unit/Db/SynchronizationContractMapperTest.php`) — asserts no INNER JOIN, single-table select on `synchronization_id`, and that the old method name is gone.
- [x] **3.2** In-scope deletion path covered by `testInScopeOrphanIsDeleted` (`tests/Unit/Service/SynchronizationServiceCleanupTest.php`).
- [x] **3.3** Cross-scope skip path covered by `testCrossScopeContractIsSkipped`.
- [x] **3.4** Missing-target skip path covered by `testMissingTargetIsSkipped`.
- [x] **3.5** Warning log on `DoesNotExistException` covered by `testWarningLoggedWhenContractLookupThrows`. Plus `testInScopeAndOutOfScopeMixedBatch` for combined-batch coverage.

### 4. Verification & follow-up audit

- [ ] **4.1** Run `composer check:strict` from `openconnector/`. **Status:** the apply environment has no `vendor/` installed (composer install not run), so the strict toolchain (PHPCS / PHPMD / Psalm / PHPStan / PHPUnit) cannot run here. Fallback: `php -l` clean for all four touched files. Run `composer install && composer check:strict` locally before merging — CI gates on PR.
- [ ] **4.2** Manually run an extern-to-intern sync against a register/schema with a known orphan and confirm: (a) the orphan's OpenRegister object is deleted, (b) the contract's `target_id` is nulled, (c) `result.objects.deleted` increments, (d) `result.timing.stages.cleanup_invalid` is present in the sync log, (e) other syncs against unrelated register/schemas are unaffected. **Status:** requires a live Nextcloud + OpenRegister environment with seed data — reserved for the user during their manual verification pass.
- [x] **4.3** `grep -rn "openregister_objects" lib/` audit. Result: zero real references; only one hit, a comment in `SynchronizationService.php:1184` explaining what was replaced. Expected outcome (zero hits) achieved.

### 5. Documentation

- [x] **5.1** Checked `docs/developers/` for cleanup-pass documentation: no matching content. `docs/synchronization-timing.md` mentions the cleanup stage at a timing level only (no implementation details), so no update needed there either.

rjzondervan:
## Summary

- Replaces the broken `INNER JOIN openregister_objects` in the synchronization cleanup query with a `synchronization_id`-only filter. The legacy table is no longer populated under OpenRegister's per-register/schema "magic table" architecture, so the JOIN had been silently returning zero rows on every sync — orphaned objects were never deleted.
- Restores the safety guard the JOIN provided by verifying each cleanup candidate is in scope via `ObjectService::find($uuid, register, schema, _rbac: false, _multitenancy: false)` before calling `deleteObject`. Out-of-scope and missing-target candidates are skipped, preventing cross-schema deletion when a sync's target was changed mid-life.
- Replaces the previously-silent `DoesNotExistException` swallow in the cleanup loop with a `LoggerInterface::warning` entry so partial failures surface.
- Adds unit tests for both the mapper rewrite and the cleanup-path scenarios (in-scope delete, cross-scope skip, missing-target skip, warning log on second lookup, mixed batch).

This PR also brings `development` up to date with `main`. The feature branch was rebased on top of a `main` merge (`cc53ce55`), so merging this into `development` will fast-forward in the 60+ commits from `main` that `development` was missing — including the `OREG 1.0.0` compatibility fix (#724), the prevent-double-wrapping fix (#714), the `enable-twig-for-endpoint` fix (#713), and several release-workflow updates.

Closes #732.

Full OpenSpec change artifacts live under `openspec/changes/fix-sync-cleanup-scope-check/` (`proposal.md`, `design.md`, `specs/synchronization-cleanup/spec.md`, `tasks.md`).

## Test plan

- [ ] `composer install && composer check:strict` from `openconnector/` (apply environment had no `vendor/` so this needs to run locally before merge — `php -l` was clean for all four touched files).
- [ ] Manual sync verification (task 4.2 in `tasks.md`): trigger an extern-to-intern sync against a register/schema with a known orphan and confirm:
  - The orphan's OpenRegister object is deleted.
  - The contract's `target_id` is nulled.
  - The sync log shows `result.objects.deleted` incremented and `result.timing.stages.cleanup_invalid` present.
  - Other syncs against unrelated `register/schema` combinations are unaffected.
- [ ] Confirm the warning log appears (and the loop continues) if `findOnTarget` ever throws `DoesNotExistException` mid-cleanup.
- [ ] CI runs against this PR (PHPCS / PHPMD / Psalm / PHPStan / unit tests).

## Follow-up (out of scope here)

- A wider audit for other JOINs against `openregister_objects` that may be silently broken in openconnector or other Conduction apps. Task 4.3 on this change confirmed openconnector's `lib/` is now clean of real references.
- An OpenRegister-side enhancement to expose a scoped `deleteObject($uuid, $register, $schema)` that refuses cross-scope, so the safety guard can live at the API boundary instead of being re-implemented in every caller.


lib/Db/RuleMapper.php
```diff
@@ -267,7 +267,7 @@ public function reorder(array $orderMap): void
 			$qb->update('openconnector_rules')
 			   ->set('order', $qb->createNamedParameter($newOrder, IQueryBuilder::PARAM_INT))
 			   ->where($qb->expr()->eq('id', $qb->createNamedParameter($ruleId, IQueryBuilder::PARAM_INT)))
-			   ->execute();
+			   ->executeQuery();
```
WilcoLouwerse: ### 🔴 Blocker — `executeQuery()` used for UPDATE — should be `executeStatement()`

`RuleMapper::reorder()` calls `->executeQuery()` on a DBAL UPDATE statement. Every other `execute()` → `executeQuery()` replacement in this PR targets SELECT queries — this one is the exception.

**Impact:** `executeQuery()` is DBAL's read path; `executeStatement()` is the write path. In DBAL 3.x the split is enforced — calling `executeQuery()` on a DML statement is undefined behaviour: may throw or silently no-op depending on the adapter. Rule reordering will break at runtime.

**Suggested fix:** Change `->executeQuery()` to `->executeStatement()` on this UPDATE. All other `executeQuery()` calls in this PR (on SELECT statements) are correct.

lib/Service/SynchronizationService.php
```diff
@@ -1123,7 +1180,40 @@ public function deleteInvalidObjects(Synchronization $synchronization, ?array $s
                     });
                 }
 
+				// Resolve OpenRegister's ObjectService for the scope-checked existence verification.
+				// This replaces the previous INNER JOIN against openregister_objects, which is no
+				// longer populated under OpenRegister's per-register/schema "magic table" architecture.
+				$openRegisterObjectService = $this->containerInterface->get('OCA\OpenRegister\Service\ObjectService');
```
WilcoLouwerse: ### 🔴 Blocker — `ContainerInterface::get()` unguarded — cleanup aborts silently if OpenRegister is unavailable

`$this->containerInterface->get('OCA\\OpenRegister\\Service\\ObjectService')` is called outside the `foreach` loop with no try/catch. If OpenRegister is not installed or the service is not registered, the exception bubbles out of `deleteInvalidObjects()` uncaught — no cleanup occurs, no log entry is written.

**Impact:** On any deployment where OpenRegister is unavailable, every sync with `targetType === 'register/schema'` silently skips all orphan cleanup. This PR's stated goal — restoring the deletion path — is defeated.

**Suggested fix:**
```php
try {
    $openRegisterObjectService = $this->containerInterface->get('OCA\\OpenRegister\\Service\\ObjectService');
} catch (\Throwable $e) {
    $this->logger->warning('deleteInvalidObjects: OpenRegister ObjectService unavailable; skipping cleanup', ['error' => $e->getMessage()]);
    break;
}
```

lib/Service/SynchronizationService.php
```diff
@@ -1123,7 +1180,40 @@ public function deleteInvalidObjects(Synchronization $synchronization, ?array $s
...
+						continue;
+					}
+
+					if ($existingObject === null) {
+						continue;
+					}
+
 					try {
 						$synchronizationContract = $this->synchronizationContractMapper->findOnTarget(synchronization: $synchronization->getId(), targetId: $targetIdToDelete);
 						if ($synchronizationContract === null) {
```
WilcoLouwerse, line 1250: ### 🔴 Blocker — `break` on null contract exits the entire `foreach` loop

The body of this `if ($synchronizationContract === null)` is `break`, which exits the `foreach ($targetIdsToDelete ...)` loop entirely. The first candidate whose contract lookup returns `null` silently aborts processing of all remaining candidates.

**Impact:** This bug was dormant while the INNER JOIN returned zero rows. This PR makes cleanup actually run — the first null-contract candidate now short-circuits the entire batch. Subsequent orphans are never cleaned up, silently.

**Suggested fix:** Change `break` to `continue` and add a log entry:
```php
if ($synchronizationContract === null) {
    $this->logger->warning('deleteInvalidObjects: no contract for target; skipping', ['targetId' => $targetIdToDelete]);
    continue;
}
```

lib/Service/SynchronizationService.php
```diff
@@ -1098,7 +1155,7 @@ public function deleteInvalidObjects(Synchronization $synchronization, ?array $s
 
 				$targetIdsToDelete = [];
 				[$registerId, $schemaId] = explode(separator: '/', string: $synchronization->getTargetId());
```
WilcoLouwerse: 🟡 **Concern — `explode()` on `getTargetId()` without null or format validation**

`[$registerId, $schemaId] = explode('/', $synchronization->getTargetId())` assumes `targetId` always contains exactly one `/`. If `getTargetId()` returns `null`, an empty string, or a plain integer (legacy format), PHP destructuring assigns `null` to `$schemaId`.

**Impact:** A null `$schemaId` passed to `ObjectService::find(..., schema: null)` may scope to all schemas, bypassing the cross-schema guard this PR specifically restores. The entire safety guarantee of the new scope check depends on this line being correct.

**Suggested fix:**
```php
$targetId = $synchronization->getTargetId();
if (!is_string($targetId) || !str_contains($targetId, '/')) {
    $this->logger->warning('deleteInvalidObjects: targetId not in register/schema format', ['targetId' => $targetId]);
    break;
}
[$registerId, $schemaId] = explode('/', $targetId, 2);
```

lib/Service/SynchronizationService.php
```diff
@@ -1123,7 +1180,40 @@ public function deleteInvalidObjects(Synchronization $synchronization, ?array $s
...
+					// the SQL-level guard this replaces did not apply RBAC either.
+					try {
+						$existingObject = $openRegisterObjectService->find(
+							id: $targetIdToDelete,
+							register: $registerId,
+							schema: $schemaId,
+							_rbac: false,
+							_multitenancy: false
+						);
+					} catch (Throwable $e) {
```
WilcoLouwerse: 🟡 **Concern — `catch (Throwable $e)` swallows logic errors and produces misleading log messages**

This catch is intentionally broad, but two issues: (1) `TypeError`, `OutOfBoundsException`, or programming errors inside `ObjectService::find` are treated identically to a legitimate scope-miss, masking dependency bugs; (2) if OpenRegister throws `DoesNotExistException` instead of returning `null` (a valid implementation), every cleaned-up orphan on the first production deploy logs a `warning` entry, flooding logs with "Scope check failed" for expected behaviour.

**Suggested fix:** Distinguish expected exceptions from unexpected ones:
```php
} catch (DoesNotExistException $e) {
    continue; // object not found — expected scope-miss
} catch (\Exception $e) {
    $this->logger->warning('Scope check failed for cleanup candidate; skipping', [
        'synchronizationId' => $synchronization->getId(),
        'targetId' => $targetIdToDelete,
        'class' => get_class($e),
        'error' => $e->getMessage(),
    ]);
    continue;
}
```

lib/Service/SynchronizationService.php
```diff
@@ -1286,7 +1383,7 @@ public function synchronizeContract(
 
         // Execute mapping if found
 		$objectBeforeMapping = $object;
-        if ($sourceTargetMapping) {
+        if ($sourceTargetMapping && $mutationType !== 'delete') {
```
WilcoLouwerse: 🟡 **Concern — `setSyncOutputOriginal()` is skipped on the delete path**

`$flowToken->setSyncOutputOriginal($object)` lives inside `if ($sourceTargetMapping && $mutationType !== 'delete')` and is therefore not called for delete mutations. The intent to skip the mapping transformation on deletes is correct — but `setSyncOutputOriginal` is for audit/tracing, not mapping, and should fire regardless of mutation type.

**Impact:** Any downstream code calling `$flowToken->getSyncOutputOriginal()` after a delete gets `null`. Delete mutations have a reduced audit trail.

**Suggested fix:** Move `setSyncOutputOriginal` before the guard:
```php
$flowToken->setSyncOutputOriginal($object);
if ($sourceTargetMapping && $mutationType !== 'delete') {
    $object = $this->mappingService->executeMapping(...);
```

lib/Service/MappingService.php
```diff
@@ -114,6 +114,20 @@ public function encodeArrayKeys(array $array, string $toReplace, string $replace
...
+     *
+     * @param string $template The Twig template to render.
+     * @param array $context The context available inside the template.
+     *
+     * @return string The rendered template result.
+     * @throws LoaderError|SyntaxError Twig exceptions
+     */
+    public function renderTemplateString(string $template, array $context = []): string
+    {
+        return html_entity_decode($this->twig->createTemplate($template)->render($context));
```
WilcoLouwerse, line 136: 🟡 **Concern — `renderTemplateString` applies `html_entity_decode` to all rendered output**

Every Twig render result is unconditionally passed through `html_entity_decode`. For endpoint URL strings this is harmless, but the method is `public` and named generically. Any caller passing a template that renders JSON or binary content will get `&lt;` → `<`, `&gt;` → `>` etc. silently injected into the output.

**Impact:** Low risk for current callers; real risk as the method's surface grows. A mapping that renders a JSON value would return silently corrupted output.

**Suggested fix:** Document the `html_entity_decode` behaviour explicitly in the docblock, or rename to `renderHtmlTemplateString` to signal the assumption. Callers that need raw output should call `$this->twig->createTemplate($template)->render($context)` directly.

tests/Unit/Service/SynchronizationServiceCleanupTest.php
```diff
@@ -0,0 +1,247 @@
...
+        return $contract;
+    }
+
+    public function testInScopeOrphanIsDeleted(): void
+    {
+        $sync = $this->makeSync();
+        $contract = $this->makeContract(self::TARGET_IN_SCOPE);
+
+        $this->contractMapper->method('findAllBySynchronization')
+            ->with(self::SYNC_ID)
```
WilcoLouwerse: 🟡 **Concern — `findAllBySynchronization` mock expects UUID string but production code passes `getId()` (integer)**

`makeSync()` sets `$sync->setId(1)` and `$sync->setUuid('sync-uuid-1')`. The production code calls `findAllBySynchronization(synchronizationId: $synchronization->getId())`, passing integer `1`. This mock's `->with(self::SYNC_ID)` expects the string `'sync-uuid-1'` — the args don't match, so PHPUnit returns `null` instead of `[$contract]`. `foreach(null ...)` throws a `TypeError` and the test fails.

The same mismatch applies to `findOnTarget(synchronization: self::SYNC_ID, ...)` on line 120 — `findOnTarget` is also called with `getId()` (integer) in production.

**Suggested fix:** Use the integer ID in the mock expectations to match production behaviour:
```php
->with($sync->getId())  // 1, not self::SYNC_ID
```
Or, if the intent is to use UUIDs as the lookup key, update the production code to call `getUuid()` instead of `getId()`.

tests/Unit/Service/SynchronizationServiceCleanupTest.php
```diff
@@ -0,0 +1,247 @@
...
+            ->willReturn($contract);
+
+        $this->openRegisterObjectService->expects($this->once())
+            ->method('find')
+            ->willReturnCallback(function (...$args) {
+                // Positional after named-arg expansion: id, _extend, files, register, schema, _rbac, _multitenancy
+                $this->assertSame(self::TARGET_IN_SCOPE, $args[0]);
+                $this->assertSame(self::REGISTER_ID, $args[3]);
+                $this->assertSame(self::SCHEMA_ID, $args[4]);
+                $this->assertFalse($args[5], '_rbac must be false to match prior unscoped JOIN behavior');
```
WilcoLouwerse: 🟡 **Concern — RBAC bypass verified by positional index — may silently pass on wrong arguments**

`$args[5]` and `$args[6]` are asserted to be `false` for `_rbac` and `_multitenancy`. PHP resolves named parameters in the callee's signature order. If `ObjectService::find`'s signature has parameters between `$schema` and `$_rbac` (e.g. `$_extend`, `$files`), the positional index is wrong and the assertion passes regardless of whether `_rbac` is actually `false`.

**Impact:** This is the only test verifying the security-critical RBAC bypass. A regression re-enabling RBAC would not be caught.

**Suggested fix:** Use `ReflectionMethod` to map parameter names to positions, or capture named args via an argument-capturing mock double:
```php
$captured = null;
$this->openRegisterObjectService->method('find')
    ->willReturnCallback(function () use (&$captured) {
        $captured = func_get_named_args(); // PHP 8 alternative or use reflection
        return $this->createMock(ObjectEntity::class);
    });
// assert $captured['_rbac'] === false after call
```

lib/Service/SynchronizationService.php
```diff
@@ -1879,9 +1987,9 @@ public function getAllObjectsFromApi(Synchronization $synchronization, ?bool $is
             usesPagination: $usesPagination
 		);
 
-		if (array_is_list($objects) === false) {
-			$objects = [$objects];
-		}
+			if ($sourceConfig['resultsPosition'] !== '_object' && array_is_list($objects) === false) {
```
WilcoLouwerse, line 2031: 🟡 **Concern — `resultsPosition _object` wrap logic is inverted between `synchronizeExternToIntern` and `getAllObjectsFromApi`**

In `synchronizeExternToIntern` (line ~634): wraps only when `resultsPosition === '_object'` and not already a list.
Here: wraps only when `resultsPosition !== '_object'`.

These are opposite behaviors for `_object` mode. If `getAllObjectsFromApi` feeds data into `synchronizeExternToIntern`, API sources using `resultsPosition: _object` may double-wrap or skip wrapping entirely.

**Impact:** Silent data corruption — objects are processed as a list of field-value pairs (unwrapped associative array) or as a double-nested array, rather than a list of one object.

**Suggested fix:** Document the intended end-to-end flow for `_object` mode and align both sites. If the intent differs between the two methods, add a comment explaining why.

lib/Service/EndpointService.php
```diff
@@ -1546,15 +1546,20 @@ private function processCustomRule(Rule $rule, array $data): array|JSONResponse
...
+    private function processWriteFileRule(Rule $rule, array $data, string $objectId, FlowToken $flowToken): array
     {
         if (isset($rule->getConfiguration()['write_file']) === false) {
             throw new Exception('No configuration found for write_file');
         }
 
         $config  = $rule->getConfiguration()['write_file'];
         $dataDot = new Dot($data);
-        $files = $dataDot[$config['filePath']];
+		$flowTokenArray = $flowToken->getRequestOriginal();
```
WilcoLouwerse: 🟢 **Minor — Mixed tabs + spaces (direct contributor to phpcs CI failure)**

The new `$flowTokenArray` / `$flowTokenDot` lines use tab indentation while the surrounding code uses 4-space indentation. This is a direct contributor to the current `phpcs` CI failure from this PR's own changes (distinguishable from pre-existing main-merge failures).

**Suggested fix:** Convert to 4-space indentation throughout the added block.

.github/workflows/release.yml
```diff
@@ -0,0 +1,36 @@
+name: Release
+
+on:
+  push:
+    branches: [main, beta, development]
```
WilcoLouwerse: 🟢 **Minor — `development` branch in trigger causes no-op workflow runs**

No job in this workflow fires on `development` branch pushes: `unstable` is `if: false`, `beta` requires `refs/heads/beta`, `stable` requires `refs/heads/main`. Every push to `development` triggers a full workflow evaluation that produces a passing-but-empty run.

**Suggested fix:** Remove `development` from the `branches` list, or re-enable `unstable` builds if that was the original intent.

WilcoLouwerse (changes requested):
3 blockers require fixes before merge: `executeQuery()` on an UPDATE in RuleMapper (should be `executeStatement()`), unguarded `ContainerInterface::get()` that aborts cleanup silently if OpenRegister is unavailable, and `break` instead of `continue` on null-contract that exits the entire cleanup loop. 7 concerns flagged (scope validation, exception handling, test correctness) — strict mode, all require attention. CI phpcs failure is partially caused by mixed indentation in this PR's own additions.

rjzondervan:
@WilcoLouwerse — review fixes pushed in `61b18d6c`. Summary of what changed and what's deferred:

## Addressed

**Blockers:**
- 🔴 `RuleMapper::reorder` — `executeQuery()` → `executeStatement()` on the UPDATE.
- 🔴 `deleteInvalidObjects` — `containerInterface->get('...ObjectService')` now wrapped in try/catch with a warning log + early `break` if OpenRegister is unavailable.

**Concerns:**
- 🟡 `deleteInvalidObjects` — `getTargetId()` is validated for `register/schema` format before `explode`. Bad format logs a warning and breaks out of the case, preventing a null `$schemaId` from silently bypassing the scope guard.
- 🟡 `deleteInvalidObjects` scope-check catch — split into `DoesNotExistException` (expected scope-miss, silent continue) and `\Exception` (logged warning with `class` name in context). No more masking of `TypeError`s as routine misses.
- 🟡 Test mock uses `$sync->getId()` (integer) for `findAllBySynchronization` and `findOnTarget` to match production behaviour.
- 🟡 RBAC bypass assertion now uses `ReflectionMethod` to map `_rbac`/`_multitenancy` names to positions — robust against `ObjectService::find` signature changes.

**Minors:**
- 🟢 `EndpointService::processWriteFileRule` — three tab-indented lines converted to 4-space.
- 🟢 `release.yml` — `development` dropped from the push trigger (no job actually runs for it).

## Not addressed (and why)

- 🔴 **Blocker on line 1219 ("`break` on null contract")** — this looks like a misread. The body of the `if ($synchronizationContract === null)` is already `continue;`. The `break;` is the `case 'register/schema': ... break;` of the switch. Happy to add a clarifying comment if you want, but I don't think there's a code change needed here.

- 🟡 **Concern 6 — `setSyncOutputOriginal` on delete path** (`SynchronizationService.php:1386`)
- 🟡 **Concern 7 — `renderTemplateString` `html_entity_decode` docblock** (`MappingService.php:128`)
- 🟡 **Concern 10 — `resultsPosition '_object'` wrap inversion** (`SynchronizationService.php:1990`)

These three target pre-existing main code unrelated to the cleanup-scope-check fix — they're in the lines this branch picked up via the `cc53ce55` merge of main, not in the apply commit. I'd suggest filing them as separate issues against main so they get their own focused PRs and proper attribution. Happy to address them here if you'd rather have everything in one PR.

Quality re-run should look better — phpcs failure on `EndpointService.php` was traced to those mixed-indent lines specifically.

lib/Db/RuleMapper.php
WilcoLouwerse: ✅ Fixed in 61b18d6: `RuleMapper::reorder` now correctly uses `executeStatement()` for the UPDATE.

lib/Service/SynchronizationService.php
WilcoLouwerse: ✅ Fixed in 61b18d6: `containerInterface->get('...ObjectService')` is now wrapped in try/catch with a warning log and early `break` if OpenRegister is unavailable.

lib/Service/SynchronizationService.php
WilcoLouwerse: ✅ Confirmed — was a misread on my part. The null-contract check already uses `continue`. The `break` I flagged is the `case 'register/schema': break` of the outer switch, which is correct.

lib/Service/SynchronizationService.php
WilcoLouwerse: ✅ Fixed in 61b18d6: `is_string()` + `str_contains('/')` guard added before `explode()`, with `limit: 2` and a warning log on bad format.

lib/Service/SynchronizationService.php
WilcoLouwerse: ✅ Fixed in 61b18d6: catch split into `DoesNotExistException` (silent continue) and `\Exception` (logged warning with class name). TypeErrors and programming errors are no longer masked as routine scope-misses.

tests/Unit/Service/SynchronizationServiceCleanupTest.php
WilcoLouwerse: ✅ Fixed in 61b18d6: `SYNC_DB_ID = 1` (integer) replaces the UUID string constant throughout; all mock `->with()` expectations now match the integer that production code passes via `getId()`.

tests/Unit/Service/SynchronizationServiceCleanupTest.php
WilcoLouwerse: ✅ Fixed in 61b18d6: `findParamPositions()` helper using `ReflectionMethod` resolves `_rbac`/`_multitenancy` by parameter name. The assertion is now robust to `ObjectService::find` signature changes. (Note: see new concern NEW-2 — the mixed-batch test uses the same pattern but hasn't been updated to use `findParamPositions()`.)

lib/Service/EndpointService.php
WilcoLouwerse: ✅ Fixed in 61b18d6: tab-indented lines in `processWriteFileRule` converted to 4-space.

.github/workflows/release.yml
WilcoLouwerse: ✅ Fixed in 61b18d6: `development` removed from the `branches` push trigger list.

lib/Service/SynchronizationService.php
WilcoLouwerse: Still open. The `&& $mutationType !== 'delete'` condition is a **new addition** in this PR's diff (line 635: the original guard was just `if ($sourceTargetMapping)`). This is not pre-existing code — it was introduced here. Please either confirm the intent to skip `setSyncOutputOriginal` on delete mutations (and add a brief code comment explaining why), or move the `setSyncOutputOriginal` call before the guard.

lib/Service/MappingService.php
WilcoLouwerse: Still open. `renderTemplateString` is a **new public method introduced in this PR** (not pre-existing code). The `html_entity_decode` applied to all Twig output is a non-obvious side effect that callers need to know about. Please add a `@note` or description line to the docblock explaining this behaviour.

lib/Service/SynchronizationService.php
WilcoLouwerse: Still open. Both the `synchronizeExternToIntern` path (array_is_list wrap guard) and the `getAllObjectsFromApi` path (resultsPosition !== '_object' condition) are **changed in this PR's diff** — neither is untouched pre-existing code. The "pre-existing, out of scope" deferral doesn't apply. Please document the intended semantics of `_object` mode in at least one of the two locations, so the asymmetry is clearly intentional rather than a latent inversion.

lib/Migration/Version1Date20250515232835.php
```diff
@@ -188,7 +188,7 @@ public function postSchemaChange(IOutput $output, Closure $schemaClosure, array
 				$query->update($tableName)
 					->set('slug', $query->createNamedParameter($update['slug']))
 					->where($query->expr()->eq('id', $query->createNamedParameter($update['id'])));
-				$query->execute();
+				$query->executeQuery();
```
WilcoLouwerse: ### 🔴 Blocker — `executeQuery()` used on UPDATE statements — should be `executeStatement()`

Two of the three `execute()` → `executeQuery()` replacements in `postSchemaChange` target UPDATE queries (this one, and the slug-clearing UPDATE around line 146). `executeQuery()` is DBAL's read path; calling it on a DML statement is undefined behaviour in DBAL 3.x and may throw or silently no-op. This is the same class of bug that was correctly fixed in `RuleMapper::reorder` (PRIOR-1), but it was reintroduced here in the migration.

**Suggested fix:** Use `executeStatement()` for both UPDATEs; leave the SELECT at line ~153 (`$result = $query->executeQuery()`) unchanged.

tests/Unit/Service/SynchronizationServiceCleanupTest.php
```diff
@@ -0,0 +1,266 @@
...
+        $sync = $this->makeSync();
+        $inScope = $this->makeContract(self::TARGET_IN_SCOPE, 'origin-a');
+        $outOfScope = $this->makeContract(self::TARGET_OUT_OF_SCOPE, 'origin-b');
+
+        $this->contractMapper->method('findAllBySynchronization')
+            ->with(self::SYNC_DB_ID)
+            ->willReturn([$inScope, $outOfScope]);
+
+        $this->openRegisterObjectService->method('find')
+            ->willReturnCallback(function ($id) {
```
WilcoLouwerse: 🟡 **Concern — `testInScopeAndOutOfScopeMixedBatch` uses positional `function ($id)` callback instead of `findParamPositions()`**

The `find()` mock callback here uses `function ($id)` (first positional arg), not the `findParamPositions()` helper introduced by the PRIOR-9 fix. If `ObjectService::find` gains a parameter before `$id`, this silently checks the wrong argument. The fix was applied to `testInScopeOrphanIsDeleted` but not carried through to this test.

**Suggested fix:** Replace `function ($id)` with a callback that uses `func_get_args()` and indexes by `$pos['id']` (same pattern as the other tests).

lib/Db/SynchronizationContractMapper.php
```diff
@@ -195,31 +195,24 @@ public function findByOriginAndTarget(string $originId, string $targetId): Synch
...
+     * in the cleanup pass via ObjectService::find, decoupling this mapper from
+     * OpenRegister's storage layout.
+     *
+     * @param string $synchronizationId The synchronization ID
      *
-     * @return array An array of target IDs or an empty array if none found
+     * @return SynchronizationContract[] An array of contracts, or [] on error
      */
-    public function findAllBySynchronizationAndSchema(string $synchronizationId, string $schemaId): array
+    public function findAllBySynchronization(string $synchronizationId): array
```
WilcoLouwerse: 🟡 **Concern — `findAllBySynchronization` declares `string $synchronizationId` but is called with an integer**

The production call site (`SynchronizationService::deleteInvalidObjects`) passes `$synchronization->getId()` which returns an integer. Non-strict PHP coerces it silently; in files with `declare(strict_types=1)` this throws a `TypeError`. The type hint and calling convention disagree.

**Suggested fix:** Either change the parameter type to `int|string $synchronizationId`, or cast at the call site: `(string) $synchronization->getId()`.

WilcoLouwerse (changes requested):
1 new blocker — [executeQuery() on UPDATE in migration](https://github.com/ConductionNL/openconnector/pull/733#discussion_r3166705476). The core scope-check fix is solid: all 3 original blockers addressed, 6 of 7 original concerns resolved, and the test quality improvements are good. 2 new concerns flagged: [positional callback in testInScopeAndOutOfScopeMixedBatch](https://github.com/ConductionNL/openconnector/pull/733#discussion_r3166705478), [string type hint vs int ID in findAllBySynchronization](https://github.com/ConductionNL/openconnector/pull/733#discussion_r3166705481). 3 prior concerns remain open (deferral as 'pre-existing' is inaccurate — all change sites are in this PR's diff): [setSyncOutputOriginal on delete path](https://github.com/ConductionNL/openconnector/pull/733#discussion_r3166696897), [renderTemplateString docblock](https://github.com/ConductionNL/openconnector/pull/733#discussion_r3166697308), [resultsPosition _object wrap asymmetry](https://github.com/ConductionNL/openconnector/pull/733#discussion_r3166697534).

rjzondervan:
@WilcoLouwerse — second pass pushed in 0978a72d. Mapping fixes to your items:

## Reopened items — now addressed

- **Reopened-1 — `setSyncOutputOriginal` on delete path** (`SynchronizationService.php:~1411`)
  Confirmed your read: the `&& $mutationType !== 'delete'` clause is new in
  this PR. Moved `setSyncOutputOriginal($object)` and `$objectBeforeMapping = $object`
  out of the guard. Mapping itself is still skipped on deletes (correct, no
  payload to transform), but the audit-trail snapshot now fires for every
  mutation. Comment added explaining the split.

- **Reopened-2 — `renderTemplateString` `html_entity_decode`**
  (`MappingService.php:128`) Added `@note` on the docblock describing the
  intentional decode for URL/path templating, and explicitly steering JSON /
  binary callers toward `$this->twig->createTemplate(...)->render(...)` directly.

- **Reopened-3 — `_object` wrap inversion** (`SynchronizationService.php:634` and `:2016`)
  Added comments at both sites explaining the asymmetry: `getAllObjectsFromApi`
  intentionally returns the single object as-is in `_object` mode (so non-sync
  callers keep the raw shape); `synchronizeExternToIntern` Stage 3 wraps it for
  iteration. The two sites are complementary, not contradictory.

## New items

- 🔴 **NEW-1 — `executeQuery()` on migration UPDATEs**
  (`Migration/Version1Date20250515232835.php`) Same DBAL read/write split as
  PRIOR-1 (`RuleMapper::reorder`). Both UPDATEs (slug back-fill and uniqueness
  UPDATE) now use `executeStatement()`; the SELECT is unchanged.

- 🟡 **NEW-2 — mixed-batch test callback**
  (`SynchronizationServiceCleanupTest.php:246`) Switched
  `testInScopeAndOutOfScopeMixedBatch` from `function ($id)` to the
  `findParamPositions()` pattern used by the other tests, so the assertion
  survives `ObjectService::find` signature changes.

- 🟡 **NEW-3 — `int|string` types on contract mapper**
  (`SynchronizationContractMapper.php`) `findAllBySynchronization` and
  `findOnTarget` now accept `int|string`. Production calls
  `Synchronization::getId()` (int); the previous `string` declarations would
  `TypeError` under `declare(strict_types=1)` and were silently coerced
  otherwise. Docblocks call out the dual-form input.

lib/Service/SynchronizationService.php
WilcoLouwerse: ✅ Addressed in 0978a72d: `setSyncOutputOriginal($object)` is now hoisted out of the `if ($sourceTargetMapping && $mutationType !== "delete")` block. Every mutation — including deletes and no-mapping passes — now leaves a pre-transformation snapshot on the flow token. The mapping execution itself remains gated correctly.

lib/Service/MappingService.php
WilcoLouwerse: ✅ Acknowledged in 0978a72d via docblock `@note`. Behavior is unchanged but the URL/path templating intent and the JSON/binary-caller restriction are now explicit on the method itself. Acceptable resolution for a 🟡 — the side effect is no longer non-obvious to future callers.

lib/Service/SynchronizationService.php
WilcoLouwerse: ✅ Acknowledged in 0978a72d. Comments at both sites (`synchronizeExternToIntern` ~L631 and `getAllObjectsFromApi` ~L2022) now explain the intentional asymmetry — the two paths are complementary, ensuring `_object` mode wraps once at the consumer and non-`_object` mode wraps once at the producer. Behavior unchanged; the rationale is now in-tree.

lib/Migration/Version1Date20250515232835.php
WilcoLouwerse: ✅ Fixed in 0978a72d: both UPDATE statements in `postSchemaChange` (slug back-fill at L146 and per-row uniqueness UPDATE at L191) now use `executeStatement()`. The intermediate SELECT at ~L153 correctly stays on `executeQuery()`. Migration parity with the `RuleMapper::reorder` fix is now complete.

tests/Unit/Service/SynchronizationServiceCleanupTest.php
WilcoLouwerse: ✅ Fixed in 0978a72d: `testInScopeAndOutOfScopeMixedBatch` switched to `function (...$args) use ($pos)` and resolves `id` via `findParamPositions()`, matching the pattern used by the other tests. The assertion is now robust to `ObjectService::find` signature changes.

lib/Db/SynchronizationContractMapper.php
WilcoLouwerse: ✅ Fixed in 0978a72d: signature widened to `int|string $synchronizationId`, with a docblock note that callers may pass either `Synchronization::getId()` (int) or a UUID string. `findOnTarget` was updated proactively in the same commit. No more silent string-coercion under `strict_types`.

WilcoLouwerse (approved):
All 6 prior findings are addressed in 0978a72d — 3 code fixes ([Migration UPDATE → executeStatement](https://github.com/ConductionNL/openconnector/pull/733#discussion_r3167745893), [mixedBatch callback](https://github.com/ConductionNL/openconnector/pull/733#discussion_r3167746322), [findAllBySynchronization signature](https://github.com/ConductionNL/openconnector/pull/733#discussion_r3167746602)) and 3 documentation-of-intent acks ([setSyncOutputOriginal hoist](https://github.com/ConductionNL/openconnector/pull/733#discussion_r3167743136), [renderTemplateString @note](https://github.com/ConductionNL/openconnector/pull/733#discussion_r3167745381), [resultsPosition asymmetry comments](https://github.com/ConductionNL/openconnector/pull/733#discussion_r3167745599)). No new findings; the new commit is small and surgical.

⚠️ Note: 8 CI checks are still failing on this branch (`lint-check`, `phpcs`, `eslint`, `stylelint`, `npm Security`, `npm License`, `SBOM`, `Deploy Documentation`). None are required by branch protection on `development`, but they should be cleaned up before merging — at minimum the PHP `phpcs` failure is worth resolving so the next merge into `main` does not inherit the noise.

PR #734, opened 2026-04-30
Fix sync cleanup: scope-checked deletion against OpenRegister magic tables

rjzondervan:
## Summary

Restores synchronization cleanup against OpenRegister 1.0+'s per-register/schema "magic table" architecture, where the existing `INNER JOIN openregister_objects` in `SynchronizationContractMapper` returned zero rows on every sync — orphaned objects were never deleted.

- `SynchronizationContractMapper`: rename `findAllBySynchronizationAndSchema` → `findAllBySynchronization`, drop the JOIN, filter contracts by `synchronization_id` only.
- `SynchronizationService::deleteInvalidObjects`: verify each candidate is in scope via `ObjectService::find($uuid, register, schema, _rbac: false, _multitenancy: false)` before calling `updateTarget('delete')`. Out-of-scope / missing-target candidates are skipped, preventing cross-schema deletion when a sync's target was changed mid-life.
- `SynchronizationService::updateTarget`: use `ObjectService::deleteObject(uuid:)` instead of the deprecated `delete(['id' => …])`.
- Replace the previously-silent `DoesNotExistException` swallow in the cleanup loop with a `LoggerInterface::warning` entry so partial failures surface.
- Unit tests for both the mapper rewrite and the cleanup-path scenarios (in-scope delete, cross-scope skip, missing-target skip, warning log on second lookup, mixed batch).
- Full OpenSpec change artifacts under `openspec/changes/fix-sync-cleanup-scope-check/` (`proposal.md`, `design.md`, `specs/synchronization-cleanup/spec.md`, `tasks.md`, `plan.json`).

Refs #732. Companion PR #733 carries the same fix plus review-feedback iterations on top, targeted at `development`.

## Test plan

- [ ] `composer install && composer check:strict` from `openconnector/`.
- [ ] Manual sync verification: trigger an extern-to-intern sync against a register/schema with a known orphan and confirm:
  - The orphan's OpenRegister object is deleted.
  - The contract's `target_id` is nulled.
  - The sync log shows `result.objects.deleted` incremented and `result.timing.stages.cleanup_invalid` present.
  - Other syncs against unrelated `register/schema` combinations are unaffected.
- [ ] Confirm the warning log appears (and the loop continues) if `findOnTarget` throws `DoesNotExistException` mid-cleanup.
- [ ] CI runs against this PR (PHPCS / PHPMD / Psalm / PHPStan / unit tests).

### DIFF
--- a/lib/Service/SynchronizationService.php
+++ b/lib/Service/SynchronizationService.php
@@ -37,6 +37,7 @@ use Psr\Container\ContainerExceptionInterface;
 use Psr\Container\ContainerInterface;
 use Psr\Container\NotFoundExceptionInterface;
 use Psr\Log\LoggerInterface;
+use Throwable;
 use React\Promise\Timer;
 use Symfony\Component\HttpKernel\Exception\TooManyRequestsHttpException;
 use Symfony\Component\Uid\Uuid;
@@ -1152,7 +1153,7 @@ class SynchronizationService
 
 				$targetIdsToDelete = [];
 				[$registerId, $schemaId] = explode(separator: '/', string: $synchronization->getTargetId());
-				$allContracts = $this->synchronizationContractMapper->findAllBySynchronizationAndSchema(synchronizationId: $synchronization->getId(), schemaId: $schemaId);
+				$allContracts = $this->synchronizationContractMapper->findAllBySynchronization(synchronizationId: $synchronization->getId());
 				$allContractTargetIds = [];
                 $allContractSourceIds = [];
 				foreach ($allContracts as $contract) {
@@ -1177,7 +1178,40 @@ class SynchronizationService
                     });
                 }
 
+				// Resolve OpenRegister's ObjectService for the scope-checked existence verification.
+				// This replaces the previous INNER JOIN against openregister_objects, which is no
+				// longer populated under OpenRegister's per-register/schema "magic table" architecture.
+				$openRegisterObjectService = $this->containerInterface->get('OCA\OpenRegister\Service\ObjectService');
+
 				foreach ($targetIdsToDelete as $targetIdToDelete) {
+					// Verify the target object exists in this sync's register/schema. If not, skip:
+					// ObjectService::deleteObject(uuid) is unscoped and would happily delete a UUID
+					// that lives in a foreign register/schema. _rbac/_multitenancy are off because
+					// the SQL-level guard this replaces did not apply RBAC either.
+					try {
+						$existingObject = $openRegisterObjectService->find(
+							id: $targetIdToDelete,
+							register: $registerId,
+							schema: $schemaId,
+							_rbac: false,
+							_multitenancy: false
+						);
+					} catch (Throwable $e) {
+						$this->logger->warning(
+							'Scope check failed for sync cleanup candidate; skipping',
+							[
+								'synchronizationId' => $synchronization->getId(),
+								'targetId' => $targetIdToDelete,
+								'error' => $e->getMessage(),
+							]
+						);
+						continue;
+					}
+
+					if ($existingObject === null) {
+						continue;
+					}
+
 					try {
 						$synchronizationContract = $this->synchronizationContractMapper->findOnTarget(synchronization: $synchronization->getId(), targetId: $targetIdToDelete);
 						if ($synchronizationContract === null) {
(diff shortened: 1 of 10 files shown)


tests/Unit/Service/SynchronizationServiceCleanupTest.php
```diff
@@ -0,0 +1,267 @@
...
+    {
+        $sync = $this->makeSync();
+        $contract = $this->makeContract(self::TARGET_OUT_OF_SCOPE);
+
+        $this->contractMapper->method('findAllBySynchronization')
+            ->with(self::SYNC_DB_ID)
+            ->willReturn([$contract]);
+
+        // Foreign-scope find returns null — object exists but lives in a different register/schema.
+        $this->openRegisterObjectService->method('find')->willReturn(null);
```
WilcoLouwerse, line 174: 🟡 **Concern — testCrossScopeContractIsSkipped does not assert find() was called with the correct register/schema**

This test stubs `ObjectService::find` to unconditionally return `null` and asserts `updateTarget` was never called — but it never verifies that `find` was actually invoked with `register: REGISTER_ID, schema: SCHEMA_ID`. A regression that passed the wrong scope (or `null`/`null`, falling back to searching across all magic tables) would still make this test pass, as long as the stubbed return is `null`.

This is the inverse of the strong assertion already present in `testInScopeOrphanIsDeleted`, which uses the `findParamPositions` reflection helper to verify all five named arguments. The cross-scope test should do the same.

**Impact:** The central security property of this PR — that cleanup cannot delete objects outside the sync's own register/schema — is not actually verified by the test that purports to cover it. A future refactor weakening the `register:`/`schema:` arguments would not be caught by CI.

**Suggested fix:** Replace the `willReturn(null)` stub with a `willReturnCallback` that asserts `$args[$pos['register']] === self::REGISTER_ID` and `$args[$pos['schema']] === self::SCHEMA_ID` before returning `null`. Apply the same fix to `testMissingTargetIsSkipped`.

lib/Service/SynchronizationService.php
```diff
@@ -1177,7 +1194,55 @@ public function deleteInvalidObjects(Synchronization $synchronization, ?array $s
...
+						'deleteInvalidObjects: OpenRegister ObjectService unavailable; skipping cleanup',
+						[
+							'synchronizationId' => $synchronization->getId(),
+							'error' => $e->getMessage(),
+						]
+					);
+					break;
+				}
+
 				foreach ($targetIdsToDelete as $targetIdToDelete) {
```
WilcoLouwerse, line 1213: 🟡 **Concern — _rbac: false / _multitenancy: false reachable from @NoAdminRequired endpoint; threat model note missing**

`deleteInvalidObjects` is invoked from `synchronize()` → `synchronizeExternToIntern()`, and `synchronize()` is reachable from `SynchronizationsController::run` and `::test`, both `@NoAdminRequired @NoCSRFRequired`. Any authenticated user who can call `/api/synchronizations-run/{id}` triggers this cleanup pass, which calls `ObjectService::find(_rbac: false, _multitenancy: false)`.

The inline comment justifies the bypass as matching the old JOIN's lack of RBAC, which is accurate. But a SQL JOIN is a storage mechanism, not an authorization layer. By porting to `ObjectService::find` with `_rbac: false`, the PR explicitly opts out of an authorization layer the modern API would otherwise apply.

The mitigating factors are real: (a) the destructive `deleteObject` call in `updateTargetOpenRegister` still defaults `_rbac: true`, so an unprivileged user cannot cause a deletion; (b) the `find` is hard-scoped to one magic table via the `register`/`schema` arguments. The exposure is UUID-existence-oracle only.

**Impact:** Low — information-disclosure only (UUID-existence oracle within the sync's own register/schema). Larger concern is the undocumented precedent of `_rbac: false` in a path reachable from `@NoAdminRequired`.

**Suggested fix:** Either (a) expand the inline comment to call out the threat model explicitly: "_rbac: false here is information-disclosure-only — the destructive path in updateTargetOpenRegister defaults _rbac=true, so no deletion bypass exists"; or (b) test whether `_rbac: true` works in the sync background context (the existing saveObject path does not appear to require the bypass) and switch to the safer default.

WilcoLouwerse (approved):
No blockers. Design is sound — the scope guard via ObjectService::find(register, schema) correctly prevents cross-schema deletion. Two concerns noted: [test coverage gap on the cross-scope guard](https://github.com/ConductionNL/openconnector/pull/734#discussion_r3168224449) and [missing threat-model note on _rbac: false reachable from @NoAdminRequired](https://github.com/ConductionNL/openconnector/pull/734#discussion_r3168224454) — address before or shortly after merge. CI failures (PHPCS/Psalm/PHPStan/PHPUnit/PHPMD) should be resolved before merging.

PR #745, opened 2026-05-07
feat: port twig endpoint templating, sync event flow fixes, and debug cleanup

bbrands02:
## Summary

This PR brings two feature branches together into development, covering synchronization event flow improvements, Twig endpoint templating, and removal of all debug output left in production code.

---

### 1. \`shouldTriggerOnEvent\` guard ported to \`handleObjectEventSynchronization\`

Previously, \`handleObjectEventSynchronization\` (triggered by OpenRegister object create/update/delete events) did not check whether a synchronization was configured to only fire on specific mutation types, causing synchronizations to run on events they were not intended for.

- Added \`shouldTriggerOnEvent()\` check to both the direct synchronization loop and the related-object trigger loop
- Synchronizations with \`sourceConfig.triggerOnlyOnEvents\` now correctly skip events not in that list
- Both loops skip already-processed synchronization IDs to prevent duplicate runs

### 2. Twig endpoint templating ported from \`main\` (commits \`d10d741\`, \`3286f01\`, \`2bab867\`)

**\`MappingService\`**
- Added public \`renderTemplateString(string \$template, array \$context): string\` — renders a Twig template string with a given context array
- Refactored the existing inline \`html_entity_decode(twig->createTemplate(...)->render(...))\` in \`executeMappingLocal\` to use the new method

**\`SynchronizationService\`**
- Added constant \`EXTRA_DATA_ENDPOINT_TEMPLATE_LOCATION = 'endpointTemplate'\`
- In \`fetchObjectProperty\`: added support for an \`endpointTemplate\` config key — the endpoint can now be a Twig template with access to \`endpoint\`, \`object\`, \`originId\`, and \`extraDataConfig\` as context variables
- In \`getAllObjectsFromApi\`: added Twig rendering for the \`endpoint\` from sourceConfig when it contains \`{{ }}\`, with \`data\` as context; rate-limit check order preserved (before endpoint extraction)

### 3. Fix: \`@self.relations\` merged into Twig context in \`getAllObjectsFromApi\`

When a synchronization rule runs at timing \`after\` on a \`register/schema\` endpoint, \`\$data\` passed to \`getAllObjectsFromApi\` is the OpenRegister response object (\`{id, @self}\`), not the original POST body. Fields like \`zaaknummer\` live inside \`@self.relations\`, making \`{{ data.zaaknummer }}\` resolve to empty.

- Relations from \`@self.relations\` are now merged into the top level of the Twig \`data\` context before rendering, so \`{{ data.zaaknummer }}\` resolves correctly from after-rule responses

### 4. \`EndpointService\` — synchronization rule flow logging

- Added \`\$this->logger->debug(...)\` tracing in \`processSyncRule\` for rule entry, fetched synchronization identity, and synchronize call parameters
- Refactored the condition/timing check in \`processRules\` to use named \`\$conditionsPassed\` and \`\$timingMatches\` variables for clarity

### 5. \`appinfo/routes.php\` — route fixes

- Removed duplicate \`dashboard#page\` route for \`/\`
- Restricted the SPA catch-all route regex to \`(?!api/).+\` to prevent it from swallowing API requests

### 6. \`EditSynchronization.vue\` — NcSelect binding fix

- Changed \`v-bind="registerOptions"\` to \`:options="registerOptions.options"\` — spreading the whole options object caused incorrect NcSelect behaviour
- Fixed \`selectedRegisterSourceValueSchemas\` and \`selectedRegisterValueSchemas\` computed properties to return \`schemas.options\` instead of \`schemas\` directly
- Applied same fixes to target register/schema selects

### 7. Removal of all debug \`echo\` / \`var_dump\` output

Removed all \`echo json_encode(...)\` and \`var_dump\` debug statements that were leaking into HTTP responses from:
- \`ObjectCreatedEventListener::handle()\`
- \`SynchronizationService::handleObjectEventSynchronization()\`
- \`SynchronizationService::synchronizeInternToExtern()\`
- \`EndpointService::processRules()\`
- \`SynchronizationService::fetchObjectProperty()\` (\`var_dump('test');die;\` that blocked execution)

---

## Test plan

- [ ] Post an object create event to the \`cloudevents/rx\` endpoint — synchronization rule fires exactly once at \`timing=after\`
- [ ] Verify \`{{ data.zaaknummer }}\` resolves in the endpoint template when sourceConfig endpoint contains \`{{ }}\`
- [ ] Confirm synchronizations with \`triggerOnlyOnEvents: [create]\` do not run on update or delete events
- [ ] Check that no debug strings appear in API responses
- [ ] Verify NcSelect register/schema dropdowns populate correctly in the EditSynchronization modal


lib/Service/SynchronizationService.php
```diff
@@ -171,19 +171,24 @@ public function handleObjectEventSynchronization(ObjectEntity $object, string $e
 
         $directSynchronizations = $this->findAllBySourceId(register: $register, schema: $schema);
         foreach ($directSynchronizations as $synchronization) {
+            if ($this->shouldTriggerOnEvent($synchronization, $eventMutationType) === false) {
```
WilcoLouwerse, line 176: 🔴 **Blocker — `shouldTriggerOnEvent()` is called but never defined on `development`.**

This call (and the matching one at line 216) will throw `Error: Call to undefined method OCA\OpenConnector\Service\SynchronizationService::shouldTriggerOnEvent()` the moment `handleObjectEventSynchronization()` is invoked, breaking every object-event synchronization.

Why: on `main` the method is defined at line 150, but on `development` it does not exist (verified — zero matches in the file, no `extends`, no traits, no `__call`). This PR ports the two _call sites_ but the diff does not include the method body.

Fix — also port the private method from `main` (lines ~150–172) into this file:

```php
/**
 * Check if a synchronization should trigger for the given object event type.
 *
 * Supported sourceConfig key:
 * - triggerOnlyOnEvents: array|string of CREATE|UPDATE|DELETE
 */
private function shouldTriggerOnEvent(Synchronization $synchronization, string $eventMutationType): bool
{
    $sourceConfig = $this->callService->applyConfigDot($synchronization->getSourceConfig());
    if (is_array($sourceConfig) === false || array_key_exists('triggerOnlyOnEvents', $sourceConfig) === false) {
        return true;
    }

    $allowedEvents = $sourceConfig['triggerOnlyOnEvents'];
    if (is_string($allowedEvents)) {
        $allowedEvents = [$allowedEvents];
    }

    if (is_array($allowedEvents) === false) {
        return true;
    }

    $allowedEvents = array_map(
        static fn ($event): string => strtoupper(trim((string) $event)),
        $allowedEvents
    );

    return in_array(strtoupper($eventMutationType), $allowedEvents, true);
}
```

lib/Service/SynchronizationService.php
```diff
@@ -208,6 +213,9 @@ public function handleObjectEventSynchronization(ObjectEntity $object, string $e
             if (in_array($synchronization->getId(), $processedSynchronizationIds, true) === true) {
                 continue;
             }
+            if ($this->shouldTriggerOnEvent($synchronization, $eventMutationType) === false) {
```
WilcoLouwerse, line 218: 🟡 **Second call site of the undefined method — same root cause as line 174.**

This is the related-object-trigger guard. It hits the same `Call to undefined method` error.

A related concern: the PR description's test plan claims all four scenarios were verified (create-only, update-only, related-object-trigger filtering, unchanged behaviour when guard returns true). With the method missing on `development`, none of those could have actually executed without throwing — either the tests were run on a branch that already had `main` merged in, or end-to-end execution was skipped. Worth re-running once the method is added so the attestation matches reality.

lib/Service/SynchronizationService.php
```diff
@@ -171,19 +171,24 @@ public function handleObjectEventSynchronization(ObjectEntity $object, string $e
 
         $directSynchronizations = $this->findAllBySourceId(register: $register, schema: $schema);
         foreach ($directSynchronizations as $synchronization) {
+            if ($this->shouldTriggerOnEvent($synchronization, $eventMutationType) === false) {
+                continue;
+            }
             try {
                 if ($eventMutationType === 'delete') {
+                    $eventObject = $object;
```
WilcoLouwerse, line 181: 🟢 **Minor — dead alias variables.**

`$eventObject = $object;` (line 179) and `$eventObjectArray = $objectArray;` (line 187) are pure aliases — neither value is transformed before being passed into `synchronize()`. They add a layer of indirection without adding clarity.

This pattern is also present on `main` so it ports faithfully, but consider passing `$object` / `$objectArray` directly to keep the diff minimal:

```php
if ($eventMutationType === 'delete') {
    $this->synchronize(
        synchronization: $synchronization,
        force: true,
        object: $object,
        mutationType: 'delete'
    );
} else {
    $this->synchronize(
        synchronization: $synchronization,
        force: true,
        object: $objectArray
    );
}
```

WilcoLouwerse (changes requested):
1 blocker — [`shouldTriggerOnEvent()` is called but never defined on `development`](https://github.com/ConductionNL/openconnector/pull/745#discussion_r3209599340), so this PR will throw `Call to undefined method` on the first object event after merge. Also a [test-plan attestation concern](https://github.com/ConductionNL/openconnector/pull/745#discussion_r3209599357) and a [minor dead-alias cleanup](https://github.com/ConductionNL/openconnector/pull/745#discussion_r3209599366). Guard placement, ordering, and dedupe logic look correct otherwise.

PR #766, opened 2026-05-16
chore(security): require 24h cooldown on npm packages

rubenvdlinde:
## Summary
- `.npmrc`: add `min-release-age=1` so `npm install` rejects any package version published less than 24h ago (npm 11.5+ native; older npm silently ignores it).
- `.github/dependabot.yml`: new file with `cooldown.default-days: 1` and `@conduction/*` on the cooldown-exclude list, so first-party Conduction releases still reach this app immediately while everything else cools down for 24h.

## Why
Recent npm supply-chain attacks (shai-hulud, nx, et al.) all rely on a window between a compromised publish and detection — typically a few hours. A blanket 24h cooldown closes that window with no functional cost, and putting it on Dependabot specifically protects against the dominant vector (bot-generated bump PRs auto-merging a poisoned version).

`@conduction/*` is excluded because we publish first-party libraries (notably `@conduction/nextcloud-vue`) actively and need fresh releases to reach this app immediately.

## Release-day escape hatch
If a contributor needs to consume a freshly-published `@conduction/*` version in a transitive dep within 24h of publish, the per-install override is:
```
npm install --min-release-age=0 @conduction/pkg@x.y.z
```

## Test plan
- [ ] `npm install` still works on this branch with no version conflicts against the existing lockfile.
- [ ] Dependabot picks up the new config on next scheduled run (no manual trigger needed).

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+updates:
+  - package-ecosystem: "npm"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 10
+    cooldown:
+      default-days: 1
+      include:
+        - "*"
+      exclude:
+        - "@conduction/*"
--- a/.npmrc
+++ b/.npmrc
@@ -1,0 +1,5 @@
+# Supply-chain hardening: reject any npm package published less than
+# 24h ago. Compromised first-party-Conduction packages are excluded via
+# Dependabot cooldown (.github/dependabot.yml); for fresh @conduction/*
+# releases, override per-install with `npm install --min-release-age=0`.
+min-release-age=1


.github/dependabot.yml
```diff
@@ -0,0 +1,13 @@
...
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    open-pull-requests-limit: 10
+    cooldown:
+      default-days: 1
+      include:
+        - "*"
+      exclude:
+        - "@conduction/*"
```
WilcoLouwerse, line 13: 🟢 **Minor — consider also covering the GitHub Actions ecosystem**

`github-actions` is a parallel supply-chain attack vector (tj-actions/changed-files, reviewdog, etc.) and benefits from the same cooldown discipline. A follow-up addition would look like:

```yaml
  - package-ecosystem: "github-actions"
    directory: "/"
    schedule:
      interval: "weekly"
    open-pull-requests-limit: 5
```

Not blocking — the npm scope here is the higher-priority vector. Fine as a follow-up if/when action pinning is on the roadmap.

WilcoLouwerse (approved):
Supply-chain hardening looks correct: `min-release-age=1` parses as 1 day in npm 11+ ([npm docs](https://docs.npmjs.com/cli/v11/using-npm/config#min-release-age)), `directory: "/"` matches `package.json` at the repo root, and the `@conduction/*` cooldown exclusion is the right shape for keeping first-party releases reachable.

CI failures present on this branch are pre-existing and unrelated to a `.npmrc` + `dependabot.yml` diff — worth a separate cleanup PR but not blocking this one.

See [`.github/dependabot.yml:13`](https://github.com/ConductionNL/openconnector/pull/766#discussion_r3258175799) for a non-blocking 🟢 Minor suggestion (extend cooldown to the `github-actions` ecosystem).

WilcoLouwerse (approved):
**Verdict: APPROVE** — npm cooldown protection is correctly configured (24h via `min-release-age=1` + Dependabot `cooldown.default-days: 1`, `@conduction/*` excluded). Optional follow-up: add a `composer` block to also cover PHP dependencies in this Nextcloud app. Pre-existing CI failures on `development` are unrelated to this 2-file config change.

**Additional findings:**

### 🟢 min-release-age unit confirmed: DAYS in npm 11.10+
npm shipped `min-release-age` in CLI v11.10.0 (Feb 2026). The unit is **days**, not milliseconds. Setting `min-release-age=1` means npm rejects any version published less than 24 h ago — exactly what the PR description claims. The comment in .npmrc ('24h ago') is accurate.

### 🟡 min-release-age requires npm ≥ 11.10.0 — CI/CD pinning should be verified
The flag was introduced in npm 11.10.0. If any CI runner or developer workstation uses an older npm (e.g. npm shipped with Node 20 LTS is ~10.x), the option is silently ignored rather than errored, offering zero protection on those machines. The PRs do not pin a Node/npm version in a `.nvmrc`, `package.json engines`, or workflow matrix. Low urgency for a chore PR, but worth a follow-up to ensure the runnner actually uses npm 11.10+.

### 🟡 Dependabot cooldown.default-days: 1 is below the community-recommended threshold
GitHub Dependabot docs and security researchers consistently recommend ≥ 7 days for meaningful protection (supply-chain attacks are typically active for 3–7 days before detection). The DataDog/StepSecurity policy scanners flag values < 7 as a security finding. Setting 1 day does provide *some* protection and aligns with the npm-side `min-release-age=1`, so it is internally consistent — but the author should be aware that raising this to 7 (or 3/7/14 per semver bump type) would be materially stronger. Not a blocker for a first iteration.

### 🟡 Dependabot cooldown.include/exclude wildcard syntax — verify @conduction/* is supported
The `cooldown` block is documented to accept wildcard patterns. GitHub's own examples show `@angular*` for scoped packages. The pattern `@conduction/*` uses a slash, which is valid glob syntax for scoped npm packages. However, official docs examples for the `exclude` list within `cooldown` (as distinct from the `ignore` top-level key) do not explicitly show the `scope/*` form with a slash; they show `@scope*`. The behavior is expected to work — Dependabot's Ruby-side pattern matching treats `/` as a literal — but it has not been verified in the diff alone. If `@conduction/*` is accidentally mis-parsed as 'any package under the `@conduction` scope with any sub-path', Conduction packages would NOT be excluded. Recommend smoke-testing after merge by checking whether Dependabot raises a PR for a `@conduction/` package within the cooldown window.

### 🟢 dependabot.yml schema is valid
Top-level `version: 2` present. The single `updates` entry contains all three required fields (`package-ecosystem`, `directory`, `schedule.interval`). The `cooldown` block with `default-days`, `include`, and `exclude` is a real, documented Dependabot option (GA announced 2025-07-01, expanded 2025-07-29). No fictional fields detected.

### 🟢 Drift between the three PRs is minimal
NV-236 and OC-766 produce byte-identical `.github/dependabot.yml` files (same git object hash `3390c9e5`). CAT-576 also produces an identical `dependabot.yml` (`3390c9e5`). The only structural difference is that CAT-576 appends to an existing `.npmrc` (lines 17+) while NV-236 and OC-766 create a new `.npmrc` from scratch — this is correct behavior for each repo. The appended block in CAT-576 is otherwise character-for-character identical to the new-file versions.

### 🟡 Dependabot cooldown only covers version updates, not security advisories
Per GitHub docs: 'The cooldown option is only available for version updates, not security updates.' This means a Dependabot security-alert PR for a patched CVE will bypass cooldown entirely — which is actually desirable behavior. The PR's stated goal (supply-chain protection for routine upgrades) is correctly served. Just make sure the team doesn't expect cooldown to gate *security* PRs as well.

### 🟡 No composer block — OC is a Nextcloud app with PHP dependencies
OpenConnector is a full-stack Nextcloud app. A `composer.json` / `composer.lock` likely exists. The `dependabot.yml` only covers `npm`; PHP dependencies managed by Composer get no Dependabot version-update PRs at all. This is pre-existing debt (Dependabot may not have been configured previously), not introduced by this PR — but since we're adding the config file now, it's a natural moment to add a `composer` block. Consider adding:
```yaml
  - package-ecosystem: "composer"
    directory: "/"
    schedule:
      interval: "weekly"
```
Not a blocker for the chore goal of this PR.

### 🟢 Pre-existing CI failures are not caused by this PR
lint-check, License, PHPcs, phpmetrics, Security, eslint, stylelint failures are all non-required and pre-exist this branch. This PR only adds two config files with no executable code; it cannot have caused or fixed those failures.

PR #787, opened 2026-05-18
feat(openconnector-adopt-or-abstractions): implement spec (#773)

rubenvdlinde:
Closes #773

## Summary

Auto-generated draft PR for OpenSpec change `openconnector-adopt-or-abstractions`.
The Hydra builder ran the spec but could not run `gh pr create` itself
(Phase D+E credential strip — Claude has no `GH_TOKEN` by design).
The entrypoint detected commits on the feature branch with no PR and
created this draft so the reviewer + security + applier can proceed.

## Spec Reference

- Issue: #773
- Spec: `/spec/`
  - `/spec/proposal.md`
  - `/spec/design.md`
  - `/spec/tasks.md`

## Commits on this branch

- 0fa574b1 chore: add task-audit.json (Rule 2a spec coherence audit) (#773)
- d9b18764 feat(tests): add bootstrap + fix phpunit.xml directory casing (#773)
- ad071a5c feat(#773): adopt OR abstractions — rename ObjectService→SourceMappingService, move constants to admin-config, declare archival/lifecycle schemas, add manifest

## Files changed

- `lib/Controller/ConsumersController.php`
- `lib/Controller/EndpointsController.php`
- `lib/Controller/EventsController.php`
- `lib/Controller/JobsController.php`
- `lib/Controller/LogsController.php`
- `lib/Controller/MappingsController.php`
- `lib/Controller/RulesController.php`
- `lib/Controller/SourcesController.php`
- `lib/Controller/SynchronizationContractsController.php`
- `lib/Controller/SynchronizationsController.php`
- `lib/EventListener/ViewDeletedEventListener.php`
- `lib/Service/CallService.php`
- `lib/Service/EndpointCacheService.php`
- `lib/Service/EndpointService.php`
- `lib/Service/ExportService.php`
- `lib/Service/ImportService.php`
- `lib/Service/JobService.php`
- `lib/Service/MappingService.php`
- `lib/Service/ObjectService.php`
- `lib/Service/RuleService.php`
- `lib/Service/SoftwareCatalogueService.php`
- `lib/Service/SourceMappingService.php`
- `lib/Service/SynchronizationService.php`
- `lib/Settings/openconnector_register.json`
- `openspec/changes/openconnector-adopt-or-abstractions/design.md`
- `openspec/changes/openconnector-adopt-or-abstractions/tasks.md`
- `openspec/manifest.yaml`
- `openspec/specs/openconnector-or-adoption/spec.md`
- `phpunit.xml`
- `task-audit.json`
- `tests/Unit/Service/EndpointCacheServiceTest.php`
- `tests/Unit/Service/SoftwareCatalogueServiceTest.php`
- `tests/Unit/Service/SourceMappingServiceTest.php`
- `tests/bootstrap.php`

---

_PR auto-created by Hydra builder entrypoint (`hydra_ensure_pr_exists`)
because Claude's session closed without running `gh pr create`.
Reviewer + applier follow as normal._


lib/Controller/JobsController.php
```diff
@@ -86,7 +86,7 @@ public function page(): TemplateResponse
      *
      * @return JSONResponse A JSON response containing the list of jobs
      */
-    public function index(ObjectService $objectService, SearchService $searchService): JSONResponse
+    public function index(SourceMappingService $objectService, SearchService $searchService): JSONResponse
```
rubenvdlinde, line 89: [unfixed: WARNING] gate-5/gate-14: Route jobs#statistics registered but no statistics() method exists. Pre-existing inherited debt.

lib/Controller/SourcesController.php
```diff
@@ -74,7 +74,7 @@ public function page(): TemplateResponse
      *
      * @return JSONResponse A JSON response containing the list of sources
      */
-    public function index(ObjectService $objectService, SearchService $searchService): JSONResponse
+    public function index(SourceMappingService $objectService, SearchService $searchService): JSONResponse
```
rubenvdlinde, line 77: [unfixed: WARNING] gate-5/gate-14: Route sources#statistics registered but no statistics() method exists. Pre-existing inherited debt.

lib/Controller/ConsumersController.php
```diff
@@ -70,7 +70,7 @@ public function page(): TemplateResponse
      *
      * @return JSONResponse A JSON response containing the list of consumers
      */
-    public function index(ObjectService $objectService, SearchService $searchService): JSONResponse
+    public function index(SourceMappingService $objectService, SearchService $searchService): JSONResponse
```
rubenvdlinde, line 73: [unfixed: WARNING] gate-7: 74 @NoAdminRequired methods across 10 in-scope controllers have no per-object auth guard. Pre-existing architectural debt — PR only substituted ObjectService→SourceMappingService type references, did not change auth logic. Adding guards to 74 methods requires new auth architecture; out of bounded fix scope.

lib/Settings/openconnector_register.json
```diff
@@ -0,0 +1,122 @@
+{
```
WilcoLouwerse, line 1: **🟡 openconnector_register.json defines lifecycle/archival annotations but no loader exists in this PR**

`lib/Settings/openconnector_register.json` declares `x-openregister-archival` and `x-openregister-lifecycle` metadata, but no PHP loader, migration step, or `SetupChecks` integration is included to actually import or activate this file at install or upgrade time. Without a loader, the annotations are declarations only and OR's archival/lifecycle runtime will never pick them up. The tasks.md notes this is gated on OR runtime shipping, but the file should either not ship yet or include a stub loader that can be activated later.

lib/Service/SynchronizationService.php
```diff
@@ -2923,7 +2989,7 @@ private function fetchFile(Source $source, string $endpoint, array $config, stri
```
WilcoLouwerse, line 2989: **🟢 findByUuid() fully replaced with find(id:) in SynchronizationService**

All four call sites that used the deprecated `$objectService->findByUuid(uuid: $objectId)` from OR's ObjectService have been replaced with `$objectService->find(id: $objectId)`. The OR `ObjectService::find()` accepts both int and string IDs and handles UUIDs correctly.

lib/Service/SynchronizationService.php
```diff
@@ -1821,15 +1875,26 @@ public function getAllObjectsFromSource(Synchronization $synchronization, ?bool
 	 */
 	public function getAllObjectsFromApi(Synchronization $synchronization, ?bool $isTest = false, ?array $data = null): array
 	{
+		// Extract source configuration
+		$sourceConfig = $this->callService->applyConfigDot($synchronization->getSourceConfig()); // TODO; This is the second time this function is called in the synchonysation flow, needs further refactoring investigation
 		//@todo this is an nuessesery db call, we should refactor this
 		$source = $this->sourceMapper->find($synchronization->getSourceId());
 
 		// Check rate limit before proceeding
```
WilcoLouwerse, line 1883: **🟢 Endpoint template rendering via Twig in getAllObjectsFromApi is safe**

The new Twig endpoint-template feature in `getAllObjectsFromApi` is gated by checking for `{{` and `}}` before evaluating, only applies to synchronisation configs (not user-facing input), and uses the existing sandboxed Twig environment. No injection vector beyond what already exists in the mapping engine.

WilcoLouwerse (approved):
**Verdict: APPROVE** — Thorough mode, no 🔴 blockers. The refactor cleanly renames `ObjectService → SourceMappingService` with a deprecated alias, replaces all `execute()` with `executeQuery()`, swaps `findByUuid()` for `find(id:)`, and moves CACHE_TTL/SUFFIX to IAppConfig. 9 🟡 concerns worth follow-up: speculative `symfony/options-resolver` + `symfony/var-exporter` deps unused; `getMapper()` 'eventSubscription' match arm unreachable after `strtolower()`; `getClient()` sanitisation is built but never applied; `openconnector_register.json` lifecycle annotations have no loader/installer; retention discrepancy (3 days code default vs `P30D` in register.json — DPO confirmation needed); `shouldTriggerOnEvent()` has no unit tests; tasks.md Phase 7 marked complete despite PROP_* constants remaining; TimedJob not migrated to ScheduledWorkflow (ADR-031, can be a separate change).

**Inline comments:**
- [🟡 openconnector_register.json defines lifecycle/archival annotations but no loader exists in this PR](https://github.com/ConductionNL/openconnector/pull/787#discussion_r3281342073)
- [🟢 findByUuid() fully replaced with find(id:) in SynchronizationService](https://github.com/ConductionNL/openconnector/pull/787#discussion_r3281343282)
- [🟢 Endpoint template rendering via Twig in getAllObjectsFromApi is safe](https://github.com/ConductionNL/openconnector/pull/787#discussion_r3281343774)

**Additional findings:**

### 🟡 getClient() copies $config to $guzzleConf but passes original $config to Guzzle
In `SourceMappingService::getClient()` (copied verbatim from the old `ObjectService`), `$guzzleConf` is created, `mongodbCluster` is unset from it, but `new Client($config)` is called with the **original** `$config` array — not `$guzzleConf`. This means `mongodbCluster` is always passed to Guzzle, which silently ignores unknown keys but the intent of the sanitisation is never fulfilled. This is a pre-existing bug that survived the rename unchanged and should be fixed to use `$guzzleConf`.

_(intended for lib/Service/SourceMappingService.php:1451)_

### 🟡 getMapper() match arm 'eventSubscription' unreachable after strtolower()
`getMapper()` downcases `$objectType` via `strtolower()` into `$objectTypeLower`, but the match arm is `'eventSubscription'` (mixed-case) rather than `'eventsubscription'`. Because PHP `match` is strict, `strtolower('eventSubscription')` === `'eventsubscription'` will fall through to the `default` throw. This bug existed in the old `ObjectService` and has been copied unchanged into `SourceMappingService`. Any caller that passes `'eventSubscription'` gets an `InvalidArgumentException` instead of the mapper.

_(intended for lib/Service/SourceMappingService.php:1682)_

### 🟡 Redundant no-op variable assignments in handleObjectEventSynchronization
Two assignments `$eventObject = $object` and `$eventObjectArray = $objectArray` are introduced immediately before those variables are passed to `synchronize()`. This adds noise without changing semantics. Likely a refactoring artifact that should be reverted or turned into a meaningful transformation.

_(intended for lib/Service/SynchronizationService.php:1801)_

### 🟡 New symfony/options-resolver and symfony/var-exporter deps added but unused in this PR
`composer.json` adds `symfony/options-resolver ^7.0` and `symfony/var-exporter ^7.2.5`, but no code in this diff or in the existing `lib/` directory imports or uses either package. They appear to be forward-looking dependencies added speculatively. This bloats the vendor bundle and can introduce transitive security debt. Remove until actually needed.

_(intended for composer.json:64)_

### 🟡 ScheduledWorkflow not adopted — TimedJob remains (ADR-031)
Per ADR-031, schedule-declarative work should go via OR `ScheduledWorkflow`, not per-app `TimedJob`. `lib/Cron/JobTask.php` still extends `OCP\BackgroundJob\TimedJob` and this PR makes no changes to the cron layer. The spec/tasks.md does not call this out as a gated task, suggesting it was not audited. This is consistent with the spec scope, but should be explicitly noted as a follow-up item.

_(intended for lib/Cron/JobTask.php:None)_

### 🟡 shouldTriggerOnEvent() new feature has no unit tests
The new `shouldTriggerOnEvent()` method adds non-trivial event-filter logic (string vs array `triggerOnlyOnEvents`, case-normalisation, early return on missing key). It is used in two call sites inside `handleObjectEventSynchronization()`. No unit test covers it — the three new test files added only cover `SourceMappingService`, `EndpointCacheService`, and `SoftwareCatalogueService`. A `SynchronizationServiceTest` testing the positive, negative, and edge cases (scalar string, array, missing key, wrong casing) would catch regressions.

_(intended for lib/Service/SynchronizationService.php:1756)_

### 🟡 SynchronizationService error-log retention discrepancy (3 days vs 30 days) left unresolved
`SynchronizationService` defaults to `259200000` ms (3 days) for error retention, while `JobService` and `CallService` use `2592000000` ms (30 days). This discrepancy is documented in `design.md` as "DPO confirmation pending" and flagged in `tasks.md`, but the PR ships without resolving it. The `x-openregister-archival` in `openconnector_register.json` declares `P30D` for **all** log schemas, creating a mismatch with the actual in-code default for sync logs. The correct value should be confirmed before the register JSON ships.

_(intended for lib/Settings/openconnector_register.json:2002)_

### 🟡 PROP_* GUID constants remain in RuleService (Phase 1 gated but tasks.md Phase 7 marked complete)
Phase 7 of tasks.md is marked `[x] Phase complete` with the note "PROP_* constants kept pending Phase 1 (RegisterResolverService gated)". However, Phase 7's spec task explicitly says "Drop eight PROP_* GUID constants". Marking Phase 7 complete while keeping the constants risks the follow-up being forgotten. The `task-audit.json` correctly flags this as a gated task, but the tasks.md checkbox is misleading and could pass a future audit as done.

_(intended for openspec/changes/openconnector-adopt-or-abstractions/tasks.md:2258)_

### 🟡 ObjectService deprecated alias correctly fires E_USER_DEPRECATED
The deprecated `ObjectService` now extends `SourceMappingService` and fires `trigger_error(..., E_USER_DEPRECATED)` in its constructor. All 11 internal call sites in lib/ have been updated to use `SourceMappingService` directly. The alias will keep external consumers working for one minor version per ADR-022.

_(intended for lib/Service/ObjectService.php:1214)_

### 🟡 executeQuery() migration is complete and correct
All deprecated `->execute()` calls in `lib/Db/` mappers and `lib/Migration/` have been replaced with `->executeQuery()`. No remaining `->execute()` calls were found in lib/.

_(intended for lib/Db/CallLogMapper.php:403)_

### 🟡 Admin-configurable TTL and suffix correctly replace hardcoded constants
`CACHE_TTL` in `EndpointCacheService` and `SUFFIX` in `SoftwareCatalogueService` are replaced with `IAppConfig` lookups that preserve current defaults (`3600` and `-sc`). `getSuffix()` is added as a public accessor for use in `RuleService`. Tests confirm constant removal and config reading.

_(intended for lib/Service/EndpointCacheService.php:673)_

### 🟡 resultsPosition='_object' double-wrap bug fixed with array_is_list guard
The fix adds `array_is_list($objectList) === false` guard before wrapping, preventing double-wrapping when the source already returns a list with `resultsPosition='_object'`. The companion fix in `getAllObjectsFromApi` also correctly skips the list-coercion for `_object` result position. Clean and minimal.

_(intended for lib/Service/SynchronizationService.php:1829)_

### 🟡 Mapping skipped for delete mutations in synchronizeContract
The `if ($sourceTargetMapping && $mutationType !== 'delete')` guard correctly prevents source→target mapping from running on delete events, which would have transformed the object being deleted and potentially caused incorrect data propagation.

_(intended for lib/Service/SynchronizationService.php:1864)_

### 🟡 MappingService::renderTemplateString extracted correctly
`html_entity_decode($this->twig->createTemplate($template)->render($context))` is correctly extracted into the public `renderTemplateString()` helper and reused in three places (mapping execution, endpoint template, extra-data endpoint template). No functional change, better reuse.

_(intended for lib/Service/MappingService.php:850)_

PR #823, opened 2026-05-20
feat(manifest-v2): LogIndex typed-primitive wrapper for /sources/logs

rubenvdlinde:
## Summary
- Introduces `src/views/wrappers/LogIndex.vue` — a thin shim over `CnIndexPage` that drives store + column selection from manifest `config.logType` (Tier-4 manifest-v2 pattern, ADR-024/036).
- Wires `/sources/logs` to LogIndex with `config: { logType: \"source\" }`; type stays `\"custom\"` until the enum grows a `\"custom-shim-over-primitive\"` value.
- Bumps manifest to v2.0.0 / `app-manifest-v2.schema.json`.
- Rewrites the `registry.js` header to document the two-flavor model: typed-primitive wrappers vs genuine bespoke views (mapping editor, cron builder, rule conditions, CloudEvent subs).

Only `logType: \"source\"` is wired up. Endpoint / Job / Synchronization / Event log routes still resolve to bespoke views — tracked as follow-ups under #814.

Refs #814

## Test plan
- [ ] Navigate to a source → Logs tab; LogIndex renders the source-log table.
- [ ] Pagination, filtering, row click work as before.
- [ ] No console errors / no regressions on Endpoint/Job/Sync/Event log routes (still bespoke).

### DIFF
--- a/src/views/wrappers/LogIndex.vue
+++ b/src/views/wrappers/LogIndex.vue
@@ -1,0 +1,174 @@
+<!-- SPDX-License-Identifier: EUPL-1.2 -->
+<!-- SPDX-FileCopyrightText: 2026 Conduction B.V. <info@conduction.nl> -->
+
+<!--
+ Generic log-index wrapper around CnIndexPage (Tier-4 manifest-v2 pattern).
+
+ Drives runtime store + column selection from `config.logType`. Manifest
+ entries reference this wrapper via `component: "LogIndex"` and pass
+ `config: { logType: "source" | "endpoint" | "job" | "synchronization" | "event" }`.
+
+ This is the larpingapp generic-wrapper pattern (a thin shim over a typed
+ primitive). Each `logType` plugs in the matching pinia store + a manual
+ column-definition list (CnIndexPage's `columns` prop instead of schema mode,
+ because openconnector log entries are not registered as OpenRegister
+ schemas yet — they come from app-owned tables).
+
+ As of this commit only `logType: "source"` is wired up. The other four
+ log routes are tracked in issue #814 as a follow-up.
+
+ @see openspec/changes — Tier-4 manifest-v2 pattern
+ @see ConductionNL/openconnector#814 — wrapper rollout to other 4 log types
+-->
+<template>
+	<CnIndexPage
+		:title="title"
+		:description="description"
+		:columns="columns"
+		:objects="rows"
+		:loading="loading"
+		:pagination="paginationProp"
+		:selectable="true"
+		:row-key="'id'"
+		@refresh="refresh"
+		@row-click="openDetail"
+		@page-changed="onPageChanged"
+		@page-size-changed="onPageSizeChanged" />
+</template>
+
+<script>
+import { CnIndexPage } from '@conduction/nextcloud-vue'
+import { logStore, navigationStore, sourceStore } from '../../store/store.js'
+
+/**
+ * Per-logType configuration: store key, fetch action, list selector.
+ * Add new entries here as more log routes adopt the wrapper.
+ */
+const LOG_TYPE_CONFIG = {
+	source: {
+		titleKey: 'Source logs',
+		descriptionKey: 'Monitor and analyze source API call logs',
+		fetch: (filters) => sourceStore.refreshSourceLogs(filters),
+		list: () => sourceStore.sourceLogs?.results ?? [],
+		total: () => sourceStore.sourceLogs?.total ?? 0,
+		modalName: 'viewSourceLog',
+	},
+}
+
+export default {
+	name: 'LogIndex',
+	components: { CnIndexPage },
+
+	props: {
+		/**
+		 * Log type discriminator. Must match a key in LOG_TYPE_CONFIG above.
+		 * Manifest-driven: pages[].config.logType.
+		 */
+		logType: {
+			type: String,
+			required: true,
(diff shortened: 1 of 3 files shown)


src/views/wrappers/LogIndex.vue
```diff
@@ -0,0 +1,174 @@
...
+		},
+	},
+
+	mounted() {
+		this.refresh()
+	},
+
+	methods: {
+		refresh() {
+			return this.config.fetch(logStore.logFilters || {})
```
WilcoLouwerse, line 157: **🔴 Pagination page/limit never sent to the API fetch call**

`onPageChanged` and `onPageSizeChanged` update `this.page` / `this.limit` and both call `this.refresh()`, but `refresh()` only passes `logStore.logFilters || {}` to `config.fetch` — the page number and page size are never forwarded to `refreshSourceLogs`. The `paginationProp` computed correctly builds `{ page, limit, total, pages }` for the UI, but clicking "Next page" will silently re-fetch page 1 every time. Either the `refresh()` call should merge `{ _page: this.page, _limit: this.limit }` into the filters, or `refreshSourceLogs` needs a `{ page, limit }` parameter.

src/views/wrappers/LogIndex.vue
```diff
@@ -0,0 +1,174 @@
...
+					formatter: (row) => {
+						const ms = row.response?.responseTime
+						return ms ? `${(ms / 1000).toFixed(3)}s` : '-'
+					},
+				},
+				{
+					key: 'created',
+					label: t('openconnector', 'Created'),
+					sortable: true,
+					formatter: (row) => row.created ? new Date(row.created).toLocaleString() : '-',
```
WilcoLouwerse, line 145: **🟡 No Vitest unit test for the new wrapper**

Every store module has a `.spec.ts` sibling. The `src/views/wrappers/` directory (new in this PR) has no corresponding spec file. Given that `LogIndex` has non-trivial logic (config discrimination, pagination prop computation, column formatters), a minimal Vitest mount test covering at least `logType` prop validation and the `paginationProp` computed would follow the project's own convention.

src/views/wrappers/LogIndex.vue
```diff
@@ -0,0 +1,174 @@
...
+	},
+
+	computed: {
+		config() {
+			return LOG_TYPE_CONFIG[this.logType]
+		},
+		title() {
+			return t('openconnector', this.config.titleKey)
+		},
+		description() {
```
WilcoLouwerse, line 88: **🟢 SPDX header present, no forbidden patterns, no PHP routes added**

The new `LogIndex.vue` carries the correct `SPDX-License-Identifier: EUPL-1.2` and `SPDX-FileCopyrightText` header. No `var_dump`, `die`, `error_log`, `print_r`, or `dd` calls appear anywhere in the diff. No PHP controller methods were added, so there are no new route-auth concerns. CI failures (lint-check, eslint, stylelint, phpcs) are all pre-existing on the `development` branch and are not caused by this PR.

WilcoLouwerse (changes requested):
**Verdict: REQUEST_CHANGES** — pagination is broken: `onPageChanged`/`onPageSizeChanged` call `refresh()` which only passes `logStore.logFilters` to the fetch; the local `this.page`/`this.limit` state is used to render the pagination UI but is never forwarded to the API, so every "next page" click silently re-fetches page 1. Fix: merge `{ _page: this.page, _limit: this.limit }` into the filter arg. Plus: manifest schema bumped to v2 (`app-manifest-v2.schema.json`, `version: 2.0.0`) but @conduction/nextcloud-vue is still `^1.0.0-beta.58` — verify v2 parser support and `config` prop injection.

**Inline comments:**
- [🔴 Pagination page/limit never sent to the API fetch call](https://github.com/ConductionNL/openconnector/pull/823#discussion_r3281344309)
- [🟡 No Vitest unit test for the new wrapper](https://github.com/ConductionNL/openconnector/pull/823#discussion_r3281344739)
- [🟢 SPDX header present, no forbidden patterns, no PHP routes added](https://github.com/ConductionNL/openconnector/pull/823#discussion_r3281344913)

**Additional findings:**

### 🟡 Manifest schema bumped to v2 but @conduction/nextcloud-vue is still ^1.0.0-beta.58
The `$schema` URL changes from `app-manifest.schema.json` (v1) to `app-manifest-v2.schema.json` and `"version"` from `"1.0.0"` to `"2.0.0"`. The installed package is `@conduction/nextcloud-vue@^1.0.0-beta.58`. It should be confirmed that the app shell in that version actually reads and validates against the v2 schema URL, and that `type: "custom"` with a nested `config` object is handled by the v2 parser. If the v2 schema is only in a newer beta, the app shell will fall back silently (the `config` prop won't be injected into `LogIndex`, making `logType` undefined and failing the `required: true` validator at runtime).

_(intended for src/manifest.json:9)_

### 🟡 `loading` computed reads `logStore.loading` but data is fetched from `sourceStore`
`refreshSourceLogs` in the source store sets `logStore.setLogsLoading(true/false)`, so `logStore.loading` does correctly reflect the source-log fetch state — this is an existing cross-store pattern. However it is fragile and non-obvious: any concurrent synchronization-log fetch would also toggle `logStore.loading`, causing a false spinner. A comment explaining the deliberate cross-store coupling would help, or consider exposing a dedicated `sourceLogsLoading` from `sourceStore`.

_(intended for src/views/wrappers/LogIndex.vue:185)_

